PR #6, opened 2022-04-10
supply block parameter to eth_call, default being BlockTag.latest

michaelsbradleyjr:
Related to https://github.com/status-im/dagger-contracts/issues/9.

When running tests against a contract deployment on a local Go Ethereum test network, I was consistently getting this error for some calls:
```json
{"code": -32602, "message": "missing value for required argument 1"}
```

The problem seemed to be related to this:

https://github.com/ethereum/go-ethereum/issues/2472
https://github.com/ethereum/go-ethereum/pull/2473

For some Ethereum clients/simulators the [default block parameter](https://eth.wiki/json-rpc/API#the-default-block-parameter) for `eth_call` is optional, but for Go Ethereum it is not.

The test suite for nim-ethers passes with my changes, and the changes allowed me to move forward with what I was attempting re: https://github.com/status-im/dagger-contracts/issues/9. However, I'm not sure I got it quite right, i.e. it may need some work before it's merge-ready.

### DIFF
--- a/ethers/contract.nim
+++ b/ethers/contract.nim
@@ -56,16 +56,20 @@ proc decodeResponse(T: type, bytes: seq[byte]): T =
     raiseContractError "unable to decode return value as " & $T
   return decoded
 
-proc call(contract: Contract, function: string, parameters: tuple) {.async.} =
+proc call(contract: Contract,
+          function: string,
+          parameters: tuple,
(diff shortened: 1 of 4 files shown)


markspanbroek:
Awesome, thanks for fixing this!

PR #10, opened 2022-05-16
Subscribe to new blocks

markspanbroek:
Provider now has a method to subscribe to new blocks. 
Equivalent to [`provider.on("block")`](https://docs.ethers.io/v5/api/providers/provider/#Provider--events) from ethers.js.


emizzle (approved):
Looks great, really well done! Much prefer your event subscription approach (utilising the type dynamism of `JsonNode`) to the way I was passing type info in the Table. Also I like the parsing of the Block type from JsonNode at the JsonRpc provider layer. 

I will rebase the confirmations feature on top of this branch. Happy for it to be merged to main, then can switch the base to main.

PR #13, opened 2022-05-23
Allow types that are equivalent to Confirmable

markspanbroek:
Allows `?TransactionResponse`, `Option[TransactionResponse]` etc to be used instead of `Confirmable`.

Figured out how to apply the solution from @Menduist (https://github.com/status-im/nim-ethers/pull/11#discussion_r875539390). Could you review @emizzle?

### DIFF
--- a/ethers/contract.nim
+++ b/ethers/contract.nim
@@ -102,35 +102,35 @@ func isConstant(procedure: NimNode): bool =
       return true
   false
 
-func isTxResponse(returntype: NimNode): bool =
-  return returntype.eqIdent($ Confirmable)
-
 func addContractCall(procedure: var NimNode) =
   let contract = procedure[3][1][0]
   let function = $basename(procedure[0])
   let parameters = getParameterTuple(procedure)
   let returntype = procedure[3][0]
-  # procedure[5] =
-  #   quote:
-  #     static: checkReturnType(type(result))
+
+  func call: NimNode =
+    if returntype.kind == nnkEmpty:
+      quote:
+        await call(`contract`, `function`, `parameters`)
+    else:
+      quote:
+        return await call(`contract`, `function`, `parameters`, `returntype`)
+
+  func send: NimNode =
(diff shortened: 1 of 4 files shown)


emizzle:
Much cleaner Mark! Regarding [@Menduist's comment](https://github.com/status-im/nim-ethers/pull/11#discussion_r875539390), he mentioned the use of `static: checkReturnType(type(result))`. I'm curious to know why that didn't work here.

emizzle (approved):

markspanbroek:
Perhaps it was added to the wrong part of the ProcDef? I saw an [assignment to `procedure[5]`](https://github.com/status-im/nim-ethers/blob/270d358b869d02a4c625dde971f799db336670fb/ethers/contract.nim#L113), but that part of the procdef is ["reserved for future use"](https://nim-lang.org/docs/macros.html#statements-procedure-declaration). Any code that should go into the body of the proc should be added to `procedure[6]`.

emizzle:
> Perhaps it was added to the wrong part of the ProcDef? I saw an [assignment to `procedure[5]`](https://github.com/status-im/nim-ethers/blob/270d358b869d02a4c625dde971f799db336670fb/ethers/contract.nim#L113), but that part of the procdef is ["reserved for future use"](https://nim-lang.org/docs/macros.html#statements-procedure-declaration). Any code that should go into the body of the proc should be added to `procedure[6]`.

Thanks for the info! Why did you not add `static:` in the latest implementation?

markspanbroek:
> Why did you not add `static:` in the latest implementation?

[I used `when`](https://github.com/status-im/nim-ethers/blob/e6c9b59cfbbcdea47f98e7f01c93e5b98ec91add/ethers/contract.nim#L125), which is already compile time, so no need for `static`.

emizzle:
Ahhh, yes of course! Thanks!

Menduist:
Even better, nice!

PR #14, opened 2022-05-23
Move `confirm` from `jsonrpc` to `provider`

markspanbroek:
Procedure `confirm` doesn't use anything JSON-RPC specific, so it can be defined on a generic `Provider`. 
It should work on all Providers (once we have more of those :)


testmodule/testJsonRpcProvider.nim
```diff
@@ -201,45 +201,42 @@ suite "JsonRpcProvider":
...
 
-    var receipt = TransactionReceipt.none
+    var receipt: TransactionReceipt
     var currentBlock = 1.u256
     var wantedConfirms = 1
     let blockHash = hexToByteArray[32](
       "0x7b00154e06fe4f27a87208eba220efb4dbc52f7429549a39a17bba2e0d98b960"
     ).some
 
-    # if TransactionReceipt is none, should not be considered mined
```
emizzle, line 211: Curious why this was removed?

emizzle (approved):
Nice work! Definitely should have been done from the start.

testmodule/testJsonRpcProvider.nim
markspanbroek: It's no longer an optional at this point. I changed the signature of `hasBeenMined` to take a plain receipt instead of an optional receipt. I check the optional receipt before calling `hasBeenMined`: https://github.com/status-im/nim-ethers/blob/c50bee2c07511d3eadb2c0a537a2a38e1f1936f2/ethers/provider.nim#L152

PR #25, opened 2022-06-29
Fix: use websock instead of news

markspanbroek:
The default was changed to news in the json_rpc
library, and broke communication with hardhat.


Menduist:
Is there a bug in news or just different functionality?

markspanbroek:
I haven't investigated it very far, I saw the CI failing with ["Connection Refused" errors](https://github.com/status-im/nim-ethers/runs/7112967506?check_suite_focus=true), and traced it to the change in json_rpc backend.

PR #27, opened 2022-07-12
Allow for custom transaction properties

markspanbroek:
Allows specifying custom transaction properties, such as `nonce` and `gasPrice` when invoking contract calls.

Example:
```nim
let overrides = TransactionOverrides(nonce: some 50.u256)
await writableToken.transfer(address, value, overrides)
```

### DIFF
--- a/ethers/contract.nim
+++ b/ethers/contract.nim
@@ -16,6 +16,11 @@ type
     provider: Provider
     signer: ?Signer
     address: Address
+  TransactionOverrides* = object
+    nonce*: ?UInt256
+    gasPrice*: ?UInt256
+    gasLimit*: ?UInt256
+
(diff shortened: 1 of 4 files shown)


emizzle (approved):
LGTM, well done.

This change would allow all `Transaction` properties to be overridden, including `sender`, `to`, `data`, and `chainId`, right? Modifying `sender`, `to` and `data` wouldn't make sense since the transaction data will be finally set in `createTransaction`. Would it make sense to create a separate type like `TransactionOverrides` that allows only `nonce`, `gasPrice`, and `gasLimit` to be updated?

markspanbroek:
Good point, in ethers.js they also check for the absence of `to` and `data` in the overrides.
I'll introduce a new type as you suggested.

Issue #23, opened 2022-06-15
How to work with private keys


JackieWeiss:
In the testmodule examples I cannot find a single example of using a private key to get and use a signer. There also doesn't appear to be any functions similar to ether.js wallet features. Is this currently not supported?

markspanbroek:
That's correct, the wallet isn't there yet.
We do accept pull requests ;)

JackieWeiss:
Just to let you know, I have started to create the Wallet functionality but it is not going to have every feature of ethers.js and it will probably take some time for me to have anything if I can even do it.
I currently do not plan to support Mnemoniks and keystores and I ran into some problems with signing Messages so I will have to see. 
I plan to use the implementation of nim-web3 to sign Transactions but I will only support Legacy and Eip1559 Transactions

markspanbroek:
That sounds great! Thanks for taking this on @JackieWeiss! Let me know if you need anything from me, I'm happy to help. Feel free to create a draft PR with your changes so far, if you feel comfortable doing so.

JackieWeiss:
So I currently have a basic wallet working. I can create Wallets from private keys, with or without a provider or with random private keys and correctly create and sign and send manual transactions. The only thing I won't be able to do alone is use the current call syntax for transactions. I am not familiar enough with macros to change the "addContractCall" function to use a different "send" function when the signer is a wallet and pass a Transaction into the "send" so it actually uses the parameters you want it to use instead of creating a default populated Transaction. For that I would require your help. I don't yet have a public repository for my changes but I will do that soon so you can look at my current changes.

markspanbroek:
Sounds like you made some nice progress! I did not anticipate any changes in the `addContractCall` macro, because it uses a `Signer` to [populate and send the transaction](https://github.com/status-im/nim-ethers/blob/440ca6360b896bd7a5f80dd70ab82cd0f9d9a1ab/ethers/contract.nim#L85). When the Wallet implements the [`Signer` interface](https://github.com/status-im/nim-ethers/blob/440ca6360b896bd7a5f80dd70ab82cd0f9d9a1ab/ethers/signer.nim) (like it does [in ethers.js](https://docs.ethers.io/v5/api/signer/#Wallet)) it should just work, but maybe I'm missing something.

JackieWeiss:
[...]

PR #28, opened 2022-07-13
Wallet

JackieWeiss:
Initial wallet and private key functionality as discussed in #23.
Can manually create and sign transactions or call functions automatically with overrides.
Some functionality is still missing but I do not currently plan to support most of them

### DIFF
--- a/ethers/wallet.nim
+++ b/ethers/wallet.nim
@@ -9,7 +9,12 @@ import ./signer
 
 export keys
 
-let rng = newRng()
+var rng {.threadvar.}: ref HmacDrbgContext
+
+proc getRng: ref HmacDrbgContext =
+  if rng.isnil:
+    rng = newRng()
+  return rng
 
 type SignableTransaction = common.Transaction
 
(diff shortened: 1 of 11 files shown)


ethers/wallet.nim
```diff
@@ -0,0 +1,140 @@
...
+    s.txType = TxLegacy
+    s.gasPrice = GasInt(tx.gasPrice.get.truncate(uint64))
+  s.chainId = ChainId(tx.chainId.get.truncate(uint64))
+  s.gasLimit = GasInt(tx.gasLimit.get.truncate(uint64))
+  s.nonce = tx.nonce.get.truncate(uint64)
+  s.to = some EthAddress(tx.to)
+  s.payload = tx.data
+  signTransaction(s, wallet.privateKey)
+
+  let rawTX = "0x" & rlp.encode(s).toHex
```
markspanbroek, line 113: You can replace the duplicated code here by calling `signTransaction` instead.

ethers/wallet.nim
```diff
@@ -0,0 +1,140 @@
...
+let rng = newRng()
+
+type SignableTransaction = common.Transaction
+
+type WalletError* = object of EthersError
+type Wallet* = ref object of Signer
+  privateKey*: PrivateKey
+  publicKey*: PublicKey
+  address*: Address
+  provider*: ?JsonRpcProvider
```
markspanbroek, line 21: Ideally, the Wallet wouldn't know about specific providers such as the `JsonRpcProvider`, only about the base `Provider` type. This would require defining `sendRawTransaction` on the base `Provider` type.

markspanbroek (approved):
Thanks @JackieWeiss, great work! I made a few comments in the code review. If you have time to address them, then please do. If not, then I'd be happy to merge your work as is, and update the code myself.

JackieWeiss:
Sorry for taking so long to look at this. Thank you for the review as well as the fixes, it really helped. I think I implemented everything you mentioned, the only exception is the teardown in testWallet.nim. I don't know why but enabling teardown produces an Illegal storage access that does not happen with the other tests. If you happen to know why this happens I would be happy to take another look

markspanbroek:
No worries, I'm happy that you took the time to make these changes :)

As for the teardown, I'm afraid you ran into [this issue](https://github.com/status-im/asynctest/pull/3) with the `asynctest` library. I just released a new version with a fix. If you update `asynctest` in `testmodule/test.nimble` to version 0.3.2, then you'll see the actual exception that triggered the illegal storage access.

Because the tests now start with a fresh EVM, the nonce needs to be 0 in every test.

markspanbroek:
Merged. Thanks for your contribution, @JackieWeiss!

PR #29, opened 2022-07-21
Fix block subscription

JackieWeiss:
Currently subscribing to new Blocks with an external provider leads to an error of 
`{"code":-32602,"message":"too many arguments, want at most 1"}`
That error does not happen with a local ganache or hardhat provider.


markspanbroek:
Looks good, thanks for fixing this! I took the liberty of cleaning it up a little bit, I hope you don't mind :)

emizzle (approved):
LGTM

PR #31, opened 2022-08-19
[feat] support distinct types

emizzle:
Add support for indexed (and non-indexed) Event fields types that are distinct `ValueType` or `SmallByteArray`. For example,
```nim
type
  DistinctAlias = distinct array[32, byte]
  MyEvent = object of Event
    a {.indexed.}: DistinctAlias
    b: DistinctAlias # also allowed for non-indexed fields
```

### Note
1. ~~This is branched off `0.2.0`, commit 16a3d25419ed849cf6791e6537ad5b3736148376, to retain compatibility with `nim-codex`. We should consider rectifying support for nim-codex. See https://github.com/status-im/nim-codex/issues/220 for more info.~~ 
EDIT: Now branched off main and will re-implement support for `nim-eth` in `nim-codex`.
2. ~~Depends on https://github.com/status-im/nim-contract-abi/pull/5, which should likely be merged first, then this PR can be updated so that it does not depend on a branch of `nim-contract-abi`.~~

### DIFF
--- a/ethers/events.nim
+++ b/ethers/events.nim
@@ -1,25 +1,10 @@
 import std/macros
-import std/typetraits
 import pkg/contractabi
 import ./basics
 import ./provider
 
 type
   Event* = object of RootObj
-  ValueType = uint8 | uint16 | uint32 | uint64 | UInt256 | UInt128 |
-              int8 | int16 | int32 | int64 | Int256 | Int128 |
-              bool | Address
-  SmallByteArray = array[ 1, byte] | array[ 2, byte] | array[ 3, byte] |
-                   array[ 4, byte] | array[ 5, byte] | array[ 6, byte] |
-                   array[ 7, byte] | array[ 8, byte] | array[ 9, byte] |
-                   array[10, byte] | array[11, byte] | array[12, byte] |
-                   array[13, byte] | array[14, byte] | array[15, byte] |
-                   array[16, byte] | array[17, byte] | array[18, byte] |
-                   array[19, byte] | array[20, byte] | array[21, byte] |
-                   array[22, byte] | array[23, byte] | array[24, byte] |
-                   array[25, byte] | array[26, byte] | array[27, byte] |
-                   array[28, byte] | array[29, byte] | array[30, byte] |
-                   array[31, byte] | array[32, byte]
 
 push: {.upraises: [].}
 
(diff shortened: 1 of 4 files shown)


markspanbroek:
Thanks @emizzle! I would love to have support for distinct types in events, but I have some doubts about this approach. The `encode()` and `decode()` for a distinct type must be explicitly supplied by the user of the library, but the check whether a distinct type fits into an indexed field is implicit. What would happen when someone defines a distinct type whose base type fits in an indexed field, but wants to encode it in more than 32 bytes or with a dynamic size?

I would like to suggest an alternative approach. We already have the `solidityType()` construction for determining the solidity type of function parameters. What if we re-used that for our check to see whether a type fits into an indexed field? So we would check whether `solidityType(T)` is `"uint8"`, `"uint16"`, etc.. We could then define default implementations for `solidityType`, `encode` and `decode` for distinct types, which the user could overwrite if necessary.

So in the simple case one could simply define:

```nim
type SimpleAlias = distinct uint64
```

and it would encode and decode as uint64 automatically.

But when you want to do something more fancy, then you can do:

```nim
type ComplexAlias = distinct uint64

func solidityType*(_: type ComplexAlias): string =
  solidityType(string)

func encode*(encoder: var AbiEncoder, value: ComplexAlias) =
  encoder.write($value)

func decode*(decoder: var AbiDecoder, T: type ComplexAlias): ?!T =
  parseInt(?decoder.read(string)).catch
```

What do you think?

emizzle:
Nice work, @markspanbroek! Even with the changes in https://github.com/status-im/nim-contract-abi/pull/5, I beleive we'd still need to [update the type check in nim-ethers to support distinct types](https://github.com/status-im/nim-ethers/pull/31/files#diff-b1383877ec76c5bea1417381bc27d5a701640685a3b0cfda29964de9452261f6R37-R56). I'll try it out.

emizzle:
Updated to depend on https://github.com/status-im/nim-contract-abi/pull/5. We should likely merge that PR first so that we don't carry over the dependency on a branch to main.

markspanbroek:
Agreed. https://github.com/status-im/nim-contract-abi/pull/5 is merged and released in 0.4.6.

emizzle:
@markspanbroek what do you think about merging this? 

https://github.com/status-im/nim-codex/pull/211 depends on both this PR and #33, so it might be easiest to merge this PR first since it's in good shape, then have that PR depend only on #33.

ethers/events.nim
```diff
@@ -33,14 +34,26 @@ func decode*[E: Event](decoder: var AbiDecoder, _: type E): ?!E =
   decoder.finishTuple()
   success event
 
+func isSupported(T: type): bool =
```
markspanbroek, line 37: Eventually I would like to see this based on solidityType(T), so that a custom type that is e.g. represented as a uint8 in solidity is also supported.

markspanbroek (approved):
Nice work!

ethers/events.nim
emizzle: `solidityType(T)` returns a `string`. How do you envision it being used here?

markspanbroek: I pushed a commit to show you what I mean. Feel free to remove it if you don't like it.

emizzle: Makes complete sense. Converting the types to strings was what I had in mind but wasn't sure if that's what you were looking for. Nice job!

PR #33, opened 2022-09-15
Add revert helpers for testing

emizzle:
Add the following helpers to help detect transaction reverts:
1. `reverts`
2. `revertsWith`
3. `doesNotRevert`
4. `doesNotRevertWith`

Catch `ValueError` from `nim-json-rpc` for now. Hopefully `nim-json-rpc` will eventually raise `JsonRpcErrors` as per https://github.com/status-im/nim-json-rpc/pull/151, at which point we can update.

### DIFF
--- a/testmodule/helpers.nim
+++ b/testmodule/helpers.nim
@@ -1,0 +1,53 @@
+import std/json
+import std/strutils
+import pkg/ethers
+
+proc revertReason*(e: ref EthersError): string =
+  try:
+    let json = parseJson(e.msg)
+    var msg = json{"message"}.getStr
+    const revertPrefixes = @[
+      # hardhat
+      "Error: VM Exception while processing transaction: reverted with " &
+      "reason string ",
+      # ganache
+      "VM Exception while processing transaction: revert "
+    ]
+    for prefix in revertPrefixes.items:
+      msg = msg.replace(prefix)
+    msg = msg.replace("\'")
+    return msg
+  except JsonParsingError:
+    return ""
+
+template reverts*(body: untyped): untyped =
+  let asyncproc = proc(): Future[bool] {.async.} =
+    try:
+      body
+      return false
+    except EthersError:
+      return true
+    except CatchableError:
+      return false
+  waitFor asyncproc()
+
+template revertsWith*(reason: string, body: untyped): untyped =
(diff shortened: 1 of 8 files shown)


testmodule/helpers.nim
```diff
@@ -0,0 +1,53 @@
...
+    return ""
+
+template reverts*(body: untyped): untyped =
+  let asyncproc = proc(): Future[bool] {.async.} =
+    try:
+      body
+      return false
+    except EthersError:
+      return true
+    except CatchableError:
```
markspanbroek, line 30: Why catch `CatchableError` here? I would expect it to be handled by the test framework like any unexpected error.

testmodule/helpers.nim
```diff
@@ -0,0 +1,53 @@
+import std/json
```
markspanbroek, line 1: If you want this to be available to users of the library, then you'd have to put in a file under `ethers/`, e.g. `ethers/testing.nim`. It can then be imported using `import pkg/ethers/testing`.

markspanbroek (approved):

testmodule/helpers.nim
emizzle: That's fair. The purpose was to keep `reverts` and `revertsWith` as booleans as implementation becomes a bit odd. Take for example this test:

```nim
  test "only checks EthersErrors":
    let ethCall = proc() =
      raise newException(ValueError, $rpcResponse)

    check:
      doesNotRevert:
        ethCall()
```

Without catching `CatchableError`, the `ValueError` is raised and needs to be caught outside of the `check doesNotRevert`, so the usage would need to become something like:
```nim
  test "only checks EthersErrors":
    let ethCall = proc() =
      raise newException(ValueError, $rpcResponse)

    var success = false

    try:
      check:
        doesNotRevert:
          ethCall()
    except ValueError:
      success = true
    check success
```
I've updated it to remove catching `CatchableError` (and also updated so that we are checking `JsonRpcProviderErrors` instead of `EthersErrors`).

testmodule/helpers.nim
emizzle: 1. Changed to `ethers/testing.nim`
2. Renamed `testmodule/testHelpers.nim` to `testmodule/testTesting.nim` and imported `pkg/ethers/testing`.

PR #35, opened 2023-02-07
support async subscription event handlers

emizzle:


markspanbroek:
Closing because we're moving towards non-async event handlers (see for instance #47).

emizzle:
Perhaps we should allow for async event handlers through an alternative handler, eg `subscribeAsync`. If there is async code that needs to be run in the handler, there is no way to avoid calling `asyncSpawn` or `waitFor`.

PR #38, opened 2023-03-24
feat: erc20 module

### LINKED ISSUE #37: Include ERC20 implementation as utils
ERC20 token is present everywhere in the ecosystem, it would be nice to include its implementation in this library for others to use as a convenience utility, so they do not have to implement it themselves.

AuHau:
Closes #37

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -143,6 +143,25 @@ await subscription.unsubscribe()
 Subscriptions are currently only supported when using a JSON RPC provider that
 is created with a websockets URL such as `ws://localhost:8545`.
 
+Utilities
+---------
+
+This library ships with some optional modules that provides convenience utilities for you such as:
+
+- `ethers/erc20` module provides you with ERC20 token implementation and its events
+
+Contribution
+------------
+
+If you want to run the tests, then before running the `nimble test`, you have to
+have installed NodeJS and then startup testing node:
+
+```shell
+$ cd testnode
+$ npm ci
+$ npm start
+```
+
 Thanks
 ------
 
--- a/Readme.md
+++ b/Readme.md
@@ -153,8 +153,8 @@ This library ships with some optional modules that provides convenience utilitie
 Contribution
 ------------
 
-If you want to run the tests, then before running the `nimble test`, you have to
-have installed NodeJS and then startup testing node:
+If you want to run the tests, then before running `nimble test`, you have to
+have installed NodeJS and started a testing node:
 
 ```shell
 $ cd testnode
--- a/ethers/erc20.nim
+++ b/ethers/erc20.nim
@@ -1,0 +1,52 @@
+import pkg/stint
+import pkg/ethers
+
+export stint
+export ethers
+
+type
+  Erc20Token* = ref object of Contract
+
+  Transfer* = object of Event
+    sender* {.indexed.}: Address
+    receiver* {.indexed.}: Address
+    value*: UInt256
+
+  Approval* = object of Event
+    owner* {.indexed.}: Address
+    spender* {.indexed.}: Address
+    value*: UInt256
+
+method name*(erc20: Erc20Token): string {.base, contract, view.}
+## Returns the name of the token.
+
+method symbol*(token: Erc20Token): string {.base, contract, view.}
+## Returns the symbol of the token, usually a shorter version of the name.
+
+method decimals*(token: Erc20Token): uint8 {.base, contract, view.}
+## Returns the number of decimals used to get its user representation.
+## For example, if `decimals` equals `2`, a balance of `505` tokens should
+## be displayed to a user as `5.05` (`505 / 10 ** 2`).
+
+method totalSupply*(erc20: Erc20Token): UInt256 {.base, contract, view.}
+## Returns the amount of tokens in existence.
+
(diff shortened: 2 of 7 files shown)


testmodule/testErc20.nim
```diff
@@ -0,0 +1,99 @@
...
+
+  test "approve tokens":
+    check (await token.balanceOf(accounts[0])) == 0.u256
+    check (await token.allowance(accounts[0], accounts[1])) == 0.u256
+
+    discard await testToken.mint(accounts[0], 100.u256)
+
+    check (await token.totalSupply()) == 100.u256
+    check (await token.balanceOf(accounts[0])) == 100.u256
+    check (await token.balanceOf(accounts[1])) == 0.u256
```
emizzle, line 70: This was tested in the test above, no need to replicate (except calling mint of course)

testmodule/testErc20.nim
```diff
@@ -0,0 +1,99 @@
...
+
+    await token.approve(accounts[1], 50.u256)
+
+    check (await token.allowance(accounts[0], accounts[1])) == 50.u256
+    check (await token.balanceOf(accounts[0])) == 100.u256
+    check (await token.balanceOf(accounts[1])) == 0.u256
+
+    await token1.transferFrom(accounts[0], accounts[1], 50.u256)
```
emizzle, line 94: When calling `approve` above, the current caller is `account[0]`, right? And so this is saying that `account[0]` allows `account[1]` to transfer 50 tokens (on behalf of `account[0]`), no?

In my mind, it should follow that we should switch to `account[1]` and then attempt to transfer `account[0]`'s tokens in `transferFrom`. I could be reading this completely wrong though 😅

testmodule/testErc20.nim
```diff
@@ -0,0 +1,99 @@
...
+  setup:
+    provider = JsonRpcProvider.new("ws://localhost:8545")
+    snapshot = await provider.send("evm_snapshot")
+    accounts = await provider.listAccounts()
+    let deployment = readDeployment()
+    testToken = TestToken.new(!deployment.address(TestToken), provider.getSigner())
+    token = Erc20Token.new(!deployment.address(TestToken), provider.getSigner())
+
+    let signer1 = provider.getSigner(accounts[1])
+    token1 = Erc20Token.new(!deployment.address(TestToken), signer1)
```
emizzle, line 34: Needed?

testmodule/testErc20.nim
```diff
@@ -0,0 +1,99 @@
...
+import pkg/ethers/erc20
+import ./hardhat
+import ./miner
+import ./mocks
+
+type
+  TestToken = ref object of Erc20Token
+
+method mint(token: TestToken, holder: Address, amount: UInt256): ?TransactionResponse {.base, contract.}
+method burn(token: TestToken, holder: Address, amount: UInt256): ?TransactionResponse {.base, contract.}
```
emizzle, line 15: Is this used? Maybe there should be an associated test?

Readme.md
```diff
@@ -143,6 +143,25 @@ await subscription.unsubscribe()
...
+
+This library ships with some optional modules that provides convenience utilities for you such as:
+
+- `ethers/erc20` module provides you with ERC20 token implementation and its events
+
+Contribution
+------------
+
+If you want to run the tests, then before running the `nimble test`, you have to
+have installed NodeJS and then startup testing node:
```
emizzle, line 157:
```suggestion
If you want to run the tests, then before running `nimble test`, you have to
have installed NodeJS and started a testing node:
```

ethers/erc20.nim
```diff
@@ -0,0 +1,52 @@
...
+    sender* {.indexed.}: Address
+    receiver* {.indexed.}: Address
+    value*: UInt256
+
+  Approval* = object of Event
+    owner* {.indexed.}: Address
+    spender* {.indexed.}: Address
+    value*: UInt256
+
+method name*(erc20: Erc20Token): string {.base, contract, view.}
```
emizzle, line 20: It might be a good idea to explicitly define default behaviour for non-overridden methods in derived classes. Typically, this is either `discard` or `raiseAssert("not implemented")`.

emizzle (approved):
Really nice Adam! A few smaller comments, but no blockers.

ethers/erc20.nim
markspanbroek: That is certainly a good idea in the general case, but here the `{.contract.}` pragma already provides the implementation.

testmodule/testErc20.nim
markspanbroek: Yes, this would probably be clearer:

```suggestion
    let signer1 = provider.getSigner(accounts[1])
    await token.connect(signer1).transferFrom(accounts[0], accounts[1], 50.u256)
```

It would also remove the need for the `token1` variable.

markspanbroek (approved):
Looks good @AuHau, thanks for adding this!
Feel free to merge after addressing the comments.

testmodule/testErc20.nim
emizzle: Ahh I see. I missed the `token1.transferFrom`. Yes, it would be clearer to have one token contract instance and switch signers as suggested.

PR #41, opened 2023-04-18
Allow contract calls to override the block tag

markspanbroek:
Allows for testing the effects of a contract call on different block heights, for instance on a previous block number, or on the pending block.


AuHau (approved):
Nice! LGTM 👍

PR #42, opened 2023-04-24
fix: pending blocks may not contains block hash

emizzle:
Pending blocks may not contain a block hash and therefore Block.hash should be optional.


markspanbroek (approved):
Good catch! Thanks!

It's a breaking change, so the next version that we release should be 0.3.0.

PR #43, opened 2023-06-09
fix: eth_call use signers address

AuHau:
Fixes issue where `msg.sender` for `eth_call` was not correctly set from the connected signer.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,7 +7,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        nim: [1.2.16, stable]
+        nim: [1.6.12, stable]
     steps:
     - name: Checkout
       uses: actions/checkout@v2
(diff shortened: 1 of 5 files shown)


markspanbroek (approved):
Nice!

Regarding the failing CI: I think we'll need to drop support for earlier versions of Nim, because we have a dependency that also dropped support (https://github.com/status-im/nim-json-rpc/pull/164), and it's not versioned.

AuHau:
> Regarding the failing CI: I think we'll need to drop support for earlier versions of Nim, because we have a dependency that also dropped support (https://github.com/status-im/nim-json-rpc/pull/164), and it's not versioned.

I don't have a problem with that, but that means that Codex will have to be also on latest. Is that a problem?

markspanbroek:
> I don't have a problem with that, but that means that Codex will have to be also on latest. Is that a problem?

It already is 1.6 only, so not a problem :smile: 
https://github.com/codex-storage/nim-codex/pull/348

PR #44, opened 2023-06-19
fix: dont export json conversions of jsonrpc

AuHau:


markspanbroek:
Looks like some of the tests check that the conversion happen correctly by calling `fromJson()`. They probably need to be moved to a separate test file that imports the conversions module.

AuHau:
Yup, hopefully this fix is alright 🙃

PR #51, opened 2023-07-14
Query past contract events

emizzle:
Based on [ethers.js's `queryFilter`](https://github.com/ethers-io/ethers.js/blob/master/packages/contracts/src.ts/index.ts#L1039), allows querying of past contract events, by querying the logs for a contract's event topic.

### DIFF
--- a/testmodule/testContracts.nim
+++ b/testmodule/testContracts.nim
@@ -213,11 +213,10 @@ for url in ["ws://localhost:8545", "http://localhost:8545"]:
 
       let currentBlock = await provider.getBlockNumber()
       let logs = await token.queryFilter(Transfer,
-                                         BlockTag.init(currentBlock - 3),
+                                         BlockTag.init(currentBlock - 1),
                                          BlockTag.latest)
 
-      check eventually logs == @[
-        Transfer(receiver: accounts[0], value: 100.u256),
+      check logs == @[
         Transfer(sender: accounts[0], receiver: accounts[1], value: 50.u256),
         Transfer(sender: accounts[1], receiver: accounts[2], value: 25.u256)
       ]
@@ -233,6 +232,6 @@ for url in ["ws://localhost:8545", "http://localhost:8545"]:
 
       let logs = await token.queryFilter(Transfer, !receipt.blockHash)
 
-      check eventually logs == @[
+      check logs == @[
         Transfer(receiver: accounts[0], value: 100.u256)
       ]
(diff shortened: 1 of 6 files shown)


testmodule/testContracts.nim
```diff
@@ -188,3 +188,51 @@ for url in ["ws://localhost:8545", "http://localhost:8545"]:
...
+    test "can query last block event log":
+
+      let signer0 = provider.getSigner(accounts[0])
+
+      discard await token.connect(signer0).mint(accounts[0], 100.u256)
+      discard await token.connect(signer0).transfer(accounts[1], 50.u256)
+
+      let logs = await token.queryFilter(Transfer)
+
+      check eventually logs == @[
```
markspanbroek, line 201: I don't think that you need `eventually` here?

```suggestion
      check logs == @[
```

testmodule/testContracts.nim
```diff
@@ -188,3 +188,51 @@ for url in ["ws://localhost:8545", "http://localhost:8545"]:
...
+      discard await token.connect(signer0).mint(accounts[0], 100.u256)
+      discard await token.connect(signer0).transfer(accounts[1], 50.u256)
+      discard await token.connect(signer1).transfer(accounts[2], 25.u256)
+
+      let currentBlock = await provider.getBlockNumber()
+      let logs = await token.queryFilter(Transfer,
+                                         BlockTag.init(currentBlock - 3),
+                                         BlockTag.latest)
+
+      check eventually logs == @[
```
markspanbroek, line 219:
```suggestion
      check logs == @[
```

testmodule/testContracts.nim
```diff
@@ -188,3 +188,51 @@ for url in ["ws://localhost:8545", "http://localhost:8545"]:
...
+      let signer0 = provider.getSigner(accounts[0])
+      let signer1 = provider.getSigner(accounts[1])
+
+      discard await token.connect(signer0).mint(accounts[0], 100.u256)
+      discard await token.connect(signer0).transfer(accounts[1], 50.u256)
+      discard await token.connect(signer1).transfer(accounts[2], 25.u256)
+
+      let currentBlock = await provider.getBlockNumber()
+      let logs = await token.queryFilter(Transfer,
+                                         BlockTag.init(currentBlock - 3),
```
markspanbroek, line 216: Wouldn't it be a more interesting test if you checked only 2 blocks back? To ensure that it takes into account the `from` block number?

markspanbroek (approved):
Very nice! Thanks @emizzle!

testmodule/testContracts.nim
emizzle: Yes, good idea. Updated.

PR #52, opened 2023-07-28
Handle invalid key when instantiating new wallet

markspanbroek:
Uses ~~an optional return value~~ Result to handle invalid key strings.

### DIFF
--- a/ethers/wallet.nim
+++ b/ethers/wallet.nim
@@ -1,4 +1,5 @@
 import eth/keys
+import ./basics
 import ./provider
 import ./transaction
 import ./signer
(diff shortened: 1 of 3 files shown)


ethers/wallet.nim
```diff
@@ -29,10 +30,16 @@ proc new*(_: type Wallet, privateKey: PrivateKey, provider: Provider): Wallet =
...
+proc new*(_: type Wallet, privateKey: string): ?Wallet =
+  if key =? PrivateKey.fromHex(privateKey):
+    some Wallet.new(key)
+  else:
+    none Wallet
+proc new*(_: type Wallet, privateKey: string, provider: Provider): ?Wallet =
+  if key =? PrivateKey.fromHex(privateKey):
+    some Wallet.new(key, provider)
+  else:
+    none Wallet
```
emizzle, line 42: Would it make more sense to return a `?!Wallet` instead? The `fromHex` error could be mapped. I think I find it odd that a constructor returns an option, which lacks context as to why, and the caller is left with a `none Wallet` and no way to know it was because of a bad private key.

emizzle (commented):
Nice work, Mark! My personally preference here would be to return `?!Wallet` from the constructor to provide error context to callers.

ethers/wallet.nim
markspanbroek: Good idea, thanks! I've changed it to return a Result instead of an Option.

emizzle (approved):
Looks great Mark, thanks for the updates!

PR #54, opened 2023-09-07
Increment nonce count when populating transaction

emizzle:
Fixes: https://github.com/codex-storage/nim-codex/issues/538

## Problem
Concurrently calling `sendTransaction` (via a contract call, for example) would call `getTransactionCount` multiple times without the underlying value being incremented before the next call.

## Solution
Keep track of a `lastSeenNonce` in the signer on each call to `populateTransaction`. On each call, the `lastSeenNonce` is compared to the `getTransactionCount` value, and when it is greater, the `lastSeenNonce` is incremented, otherwise it is set to the `getTransactionCount` value.

Additionally, if a transaction is crated manually and sent (via `sendTransaction`), the `lastSeenNonce` is updated.


markspanbroek (approved):
Very nice, thanks @emizzle!

PR #55, opened 2023-09-27
prevent stuck transactions by async locking nonce sequencing (+ estimate gas)

emizzle:
During `populateTransaction`, a new nonce is obtained using `signer.getNonce`. After this, the transaction's gas is estimated, often times resulting in a failure because the simulated transaction failed (due to a contract revert, nonce too high/low, gas error). The next transaction created will have a nonce that is one too high, resulting in that transaction getting stuck until the previous transaction is executed. To solve this, the previous transaction can be cancelled by sending a 0-value transaction to the sender of the transaction with the same nonce as the previous transaction. This will ensure that any queued transactions with higher nonces will be able to be executed.


AuHau (commented):
This also LGTM, but again I would live approve to @markspanbroek

ethers/contract.nim
```diff
@@ -116,14 +117,46 @@ proc call(contract: Contract,
...
+      return await signer.populateTransaction(tx)
+    except EstimateGasError as e:
+      if nonce =? e.transaction.nonce:
+
+        if lastSeenNonce =? signer.lastSeenNonce and
+          lastSeenNonce > nonce:
+            discard await signer.cancelTransaction(e.transaction)
+            let revertReason = if not e.parent.isNil: e.parent.msg
+                               else: "unknown"
+            info "A cancellation transaction has been sent to prevent stuck " &
```
markspanbroek, line 136: Chronicles is not currently a dependency of nim-ethers (and I'd prefer it if we keep it that way :)

ethers/contract.nim
```diff
@@ -116,14 +117,46 @@ proc call(contract: Contract,
   let response = await contract.provider.call(transaction, overrides)
   return decodeResponse(ReturnType, returnMultiple, response)
 
+proc populateTransaction(
```
markspanbroek, line 120: I would expect the logic to handle stuck transactions to be at the level of the signer, not the level of the contract invocations? Stuck transactions can happen also when using the signer directly, I think?

ethers/signer.nim
```diff
@@ -44,26 +52,62 @@ method estimateGas*(signer: Signer,
...
+  # with the failed tx's nonce
+
+  without sender =? tx.sender:
+    raiseSignerError "transaction must have sender"
+  if sender != await signer.getAddress():
+    raiseSignerError "can only cancel a tx this signer has sent"
+  without nonce =? tx.nonce:
+    raiseSignerError "transaction must have nonce"
+
+  var cancelTx = tx
```
markspanbroek, line 97: Should we clear the tx.data as well? And what about fees etc..?

ethers/testing.nim
```diff
@@ -32,5 +33,12 @@ proc reverts*[T](call: Future[T], reason: string): Future[bool] {.async.} =
     else:
       discard await call
     return false
-  except ProviderError as error:
-    return reason == error.revertReason
+  except ProviderError, SignerError:
+    let error = getCurrentException()
```
markspanbroek, line 37: Doesn't this work?

```suggestion
  except ProviderError, SignerError as error:
```

markspanbroek (commented):
I'd like to avoid sending transactions that were not specifically asked for by the user of the library. To solve the problem of stuck transactions, I think we should either:

1. ensure that the situation never occurs
2. signal when this occurs, and let the user of the library decide what to do with it (e.g. cancel the transaction)

If we go for option (2), then we could introduce a specific error that is only raised when a transaction will get stuck, and expose a means of handling it (e.g. the cancelTransaction() proc from this PR)

But thinking about this a bit more, we might actually be able to pull off option (1). A transaction can get stuck if there are multiple concurrent invocations of `populateTransaction`, whereby the nonce is increased a couple of times, but one of the invocations fails on `estimateGas`. If we ensure that we handle a single `populateTransaction` at a time, and never do this concurrently, then I think we will not run into this problem anymore. Either an AsyncLock or an AsyncQueue can probably be used to ensure that increasing the nonce and estimating gas either both succeed, or both fail.

In pseudocode:
```nim
try:
  await lock.acquire()
  populated.nonce = await signer.getNonce()
  populated.gasLimit = await signer.estimateGas()
except EstimateGasError:
  signer.decreaseNonce()
  raise
finally:
  await lock.release()
```

ethers/testing.nim
emizzle: It does not

emizzle:
I'm not sure if that will solve the issue because the problem occurs when populateTransaction is called multiple times but the transactions are not sent or failed to send. Adding an async lock would prevent concurrency in populating the transaction (and nonce), but it would not wait for the transaction to be sent, nor do I think that's something we want.

I think we can come up with a version of #2, where something like a `StuckTransactionError` is raised in `sendTransaction` which would allow the user to call `cancelTransaction`. Alternatively or in addition to, we could add a parameter to `sendTransaction` that enables cancelling potentially stuck transactions instead automatically.

ethers/contract.nim
emizzle: I thought about this too, but ultimately I feel that almost every transaction that may run into this issue originates from a contract transactions, and any others would have to be initiated from a user creating a transaction and submitting it through the signer, in which case they would get the `EstimateGasError` and they can handle the error as they see fit, including calling `cancelTransction` (which is on the level of the signer).

ethers/contract.nim
emizzle: After our discussion about removing chronicles, I've removed it from this PR, but it is still in #57.

ethers/contract.nim
emizzle: This logic has been removed.

ethers/contract.nim
```diff
@@ -3,11 +3,13 @@ import std/macros
 import std/sequtils
 import pkg/chronos
 import pkg/contractabi
+import pkg/stew/byteutils
 import ./basics
 import ./provider
 import ./signer
 import ./events
 import ./fields
+import ./exceptions
```
markspanbroek, line 12: Are these imports still needed? I see no changes anymore in this file?

ethers/exceptions.nim
```diff
@@ -0,0 +1,7 @@
+import ./basics
+
+func msgStack*(error: ref EthersError): string =
```
markspanbroek, line 3: Doesn't seem to be used anymore?

ethers/providers/jsonrpc.nim
```diff
@@ -3,6 +3,7 @@ import std/tables
 import std/uri
 import pkg/json_rpc/rpcclient
 import pkg/json_rpc/errors
+import pkg/stew/byteutils
```
markspanbroek, line 6: Unused import?

ethers/signer.nim
```diff
@@ -1,16 +1,35 @@
 import ./basics
 import ./provider
+import pkg/chronicles
 
 export basics
+export chronicles
+
+logScope:
+  topics = "ethers signer"
```
markspanbroek, line 9: There's no logging anymore in this PR, so this can probably also be removed?

ethers/signer.nim
```diff
@@ -64,6 +87,39 @@ method updateNonce*(signer: Signer, nonce: ?UInt256) {.base, gcsafe.} =
...
+
+proc ensureNonceSequence(
+  signer: Signer,
+  tx: Transaction
+): Future[Transaction] {.async.} =
+  ## Ensures that once the nonce is incremented, if estimate gas fails, the
+  ## nonce is decremented. Disallows concurrent calls by use of AsyncLock.
+  ## Immediately returns if tx already has a nonce or gasLimit.
+
+  if not (tx.nonce.isNone and tx.gasLimit.isNone):
```
markspanbroek, line 102: Although equivalent, this reads a little better, I think:

```suggestion
  if tx.nonce.isSome or tx.gasLimit.isSome:
```

ethers/signer.nim
```diff
@@ -73,17 +129,33 @@ method populateTransaction*(signer: Signer,
...
     populated.chainId = some(await signer.getChainId())
-  if transaction.gasPrice.isNone and (transaction.maxFee.isNone or transaction.maxPriorityFee.isNone):
+  if populated.gasPrice.isNone and (populated.maxFee.isNone or populated.maxPriorityFee.isNone):
     populated.gasPrice = some(await signer.getGasPrice())
-  if transaction.gasLimit.isNone:
-    populated.gasLimit = some(await signer.estimateGas(populated))
+  if populated.nonce.isNone:
+    populated.nonce = some(await signer.getNonce())
+  if populated.gasLimit.isNone:
+          populated.gasLimit = some(await signer.estimateGas(populated))
```
markspanbroek, line 143: It feels a bit weird that we're estimating gas before setting all properties, such as gasLimit and sender.
I would suggest to either do the call to `ensureNonceSequence` last, or to combine `populateTransaction` and `ensureNonceSequence` into a single method.

markspanbroek (approved):
Thanks @emizzle, this is great!
I have a few small remarks, but otherwise good to go!

ethers/signer.nim
emizzle: Fair point. I've merged `ensureNonceSequence` into `populateTransaction`.

ethers/signer.nim
```diff
@@ -129,18 +95,40 @@ method populateTransaction*(signer: Signer,
...
+    # stuck transactions
+    try:
+      populated.nonce = some(await signer.getNonce())
+      populated.gasLimit = some(await signer.estimateGas(populated))
+    except ProviderError, EstimateGasError:
+      let e = getCurrentException()
+      signer.decreaseNonce()
+      raise e
+    finally:
+      signer.populateLock.release()
```
markspanbroek, line 125: The `try:finally` should be around all the statements following `lock.acquire()`. If one of the `getAddress()` or `getChainId()` or ... fails, then the lock is stuck forever.

emizzle: https://github.com/codex-storage/nim-ethers/pull/58

ethers/signer.nim
```diff
@@ -73,17 +95,55 @@ method populateTransaction*(signer: Signer,
   if chainId =? transaction.chainId and chainId != await signer.getChainId():
     raiseSignerError("chain id mismatch")
 
+  if signer.populateLock.isNil:
```
dryajov, line 98: You want to initialize this in the constructor.

dryajov: What happened to this @emizzle ?

PR #57, opened 2023-10-18
On transaction failure, fetch revert reason with replayed transaction

emizzle:
When calling `.sendTransaction`, the flow is:
1. populate transaction with values, including gasLimit using `estimateGas`
2. send transaction

In step 1, if there is an error during gas estimation, an `EstimateGasError` is raised with the revert reason.

In step 2, if the transaction receipt is returned with a status of failed (requires the transaction to be confirmed), the transaction is replayed at the previous block (the full state transitions of the *previous* block are including, and the state transitions from earlier txs of the *same* block are *not* included), and a revert reason is obtained from the replay. A `ProviderError` is raised with the revert reason.

The replay flow is:
1. Get transaction at TransactionHash
2. Deserialize that transaction into a `PastTransaction` type
3. Convert `PastTransaction` to a `Transaction`
4. Simulate the transaction at a block number by *calling* `eth_sendTransaction`, and providing the transaction object and a block number.

### NOTE:
1. This PR is based on top of #55
2. nim-chronicles was added in, with some discussion around not including it due to it not being versioned. It does, however appear to be versioned, but I'm not entirely sure if this is enough. The other option, which would suit a future PR, is to add chronicles as an optional dependency.


AuHau (commented):
It does look LGTM, but honestly I don't have that good understanding here so I would leave approval to @markspanbroek

markspanbroek (commented):
This is great, thanks @emizzle!

I think I'd like review this again after the comments in #55 have been addressed. Some of them apply to this PR as well.

emizzle:
Ready for another review 👍

emizzle:
Note: nim-chronicles was added in this PR (originally was in #55), with some discussion around not including it due to it not being versioned. It does, however appear to be versioned, but I'm not entirely sure if this is enough. The other option, which would suit a future PR, is to add chronicles as an optional dependency.

ethers/testing.nim
```diff
@@ -1,4 +1,5 @@
 import std/strutils
+import ./contract
```
markspanbroek, line 2: Unused import?

testmodule/testContracts.nim
```diff
@@ -5,6 +5,7 @@ import pkg/questionable
 import pkg/stint
 import pkg/ethers
 import pkg/ethers/erc20
+import pkg/ethers/wallet
```
markspanbroek, line 8: Unused import?

markspanbroek (approved):
Looks great @emizzle! This is going to help us a lot in debugging transactions.

PR #58, opened 2023-10-26
[fix] Ensure AsyncLock is released in case of exception

emizzle:
Wrap try/finally around populateTransaction logic to ensure the lock is always released in the case of an error

### DIFF
--- a/ethers/signer.nim
+++ b/ethers/signer.nim
@@ -103,33 +103,35 @@ method populateTransaction*(signer: Signer,
 
   var populated = transaction
 
-  if transaction.sender.isNone:
-    populated.sender = some(await signer.getAddress())
-  if transaction.chainId.isNone:
-    populated.chainId = some(await signer.getChainId())
-  if transaction.gasPrice.isNone and (transaction.maxFee.isNone or transaction.maxPriorityFee.isNone):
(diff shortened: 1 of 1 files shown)


ethers/signer.nim
```diff
@@ -103,33 +103,35 @@ method populateTransaction*(signer: Signer,
...
+    if transaction.gasPrice.isNone and (transaction.maxFee.isNone or transaction.maxPriorityFee.isNone):
+      populated.gasPrice = some(await signer.getGasPrice())
+
+    if transaction.nonce.isNone and transaction.gasLimit.isNone:
+      # when both nonce and gasLimit are not populated, we must ensure getNonce is
+      # followed by an estimateGas so we can determine if there was an error. If
+      # there is an error, the nonce must be deprecated to prevent nonce gaps and
+      # stuck transactions
+      try:
+        populated.nonce = some(await signer.getNonce())
```
markspanbroek, line 120: When `getNonce()` fails, you don't want to decrease the nonce, so I'd move it out of the `try:except` block:

```suggestion
      populated.nonce = some(await signer.getNonce())
      try:
```

markspanbroek (approved):

PR #61, opened 2023-12-15
Updates for Chronos V4

gmega:
This PR adds minimal adjustments to ensure that nim-ethers works (i.e. compiles and the test suite passes) with Chronos V4. Changes are minimal enough that it should still work with Chronos 3.X, though that's more of a bonus than a goal. 

Two of the more controversial points in this PR are:

  * the `{.cast(raises: []).}` in table operations which 1.6.x will otherwise (erroneously) declare as raisers of Exception. I had asked for this in nim-forum with a smaller example and Araq suggested using `{.experimental: "strictEffects".}` (https://forum.nim-lang.org/t/10749) which indeed fixed my example, but for reasons I can't understand it doesn't work here.
  * the bump to nim 1.6.16, which I had to do to work around what appeared to be memory getting corrupted with 1.6.14. I did a lot of digging into Chronos and couldn't really find an issue there, so I'm chalking this off as a compiler issue.


ethers/signer.nim
```diff
@@ -26,10 +26,12 @@ proc raiseEstimateGasError(
     parent: parent)
   raise e
 
-method provider*(signer: Signer): Provider {.base, gcsafe.} =
+method provider*(signer: Signer):
+    Provider {.base, gcsafe, raises: [CatchableError].} =
```
dryajov, line 30: Hmm... I would not blanked raise `CatchableError`, in fact I would set this to `raises: []` and only list those exceptions that are intended to propagate. The whole point of this is preventing spurious exceptions from bubbling up, otherwise there is no point in tracking exceptions at all.

dryajov: In fact I would use the `push: raises []` pragma in all the files and only override this per method on an as needed basis.

gmega: Yeah, I mean, as I was working through this, I was faced with two choices:

  1. do the minimal amount of changes so that the current implementation works with Chronos V4;
  2. actually track the entire exception chains and expose the minimal set of exceptions that all method implementations may throw.

I went with 1 to get there faster and make the changes smaller. But we can definitely do 2.

gmega: Yeah, I mean, as I was working through this, I was faced with two choices:

  1. do the minimal amount of changes so that the current implementation works with Chronos V4;
  2. actually track the entire exception chains and expose the minimal set of exceptions that all method implementations may throw, which is what you're proposing.

(2) seems like a good idea, but it is actually sort of tricky. Without some feeling as to how the API has to be designed, I could easily replace `CatchableError` with `EthersError` pretty much everywhere, and it would compile without much gain in exception semantics. The actual narrow types for the exceptions that would make sense to expose depends on what we expect clients to want to try to recover from, so that's why I ended up going with the (1) route.

I'll give it a stab anyhow, let's see what comes out of it.

ethers/providers/jsonrpc/subscriptions.nim
```diff
@@ -14,6 +14,14 @@ type
     callbacks: Table[JsonNode, SubscriptionCallback]
   SubscriptionCallback = proc(id, arguments: JsonNode) {.gcsafe, raises:[].}
 
+# FIXME Nim 1.6.XX seems to have issues tracking exception effects and will see
```
dryajov, line 17: Actually, this is probably the wrong way of doing it, as you noted in the forum post, annotating the callback type with a `raises: []` fixes the probable, this is in fact the correct fix, callbacks should (almost) never allow exceptions to escape.

ethers/signer.nim
dryajov: Yeah, I understand, this gets messy pretty quickly, but I still think that push pragma with `raises: []` and an explicit `EthersError` where required is a much better than the bare `CatchableError`. Also, I suspect that a lot of issues are stemming from the fact that callbacks aren't annotated with `raises: []`.

ethers/providers/jsonrpc/subscriptions.nim
dryajov: Also, here is a pointer on how to handle the strict effect changes in a backwards/forwards compatible way in the 1.6.0 release notes - https://github.com/narimiran/Nim/blob/f14925d3bc778bcb7eef4eb9085f08d65f4281ce/changelogs/changelog_1_6_0.md#strict-effects, should be helpful.

emizzle: In summary, the solution would be a combination of using `raises:[]` in the callback definition, and changing the callback table key type.

emizzle (commented):
Exception handling is always super tricky!

ethers/providers/jsonrpc/subscriptions.nim
dryajov: Note that table accessors such as `get` or `[]` will raise an exception, so this should either use the non throwing versions (getOrDefault, etc), or handle the exceptions in the callback itself.

gmega: I love when two different issues conflate into the same symptom. :-D I'm not sure I follow your explanation though @emizzle (and that's probably the problem :-)): is parseJson called during the hashing of JsonNode?

@dryajov yes but in this case we're writing to the table (`[]=`), this should not raise exceptions.

ethers.nimble
```diff
@@ -3,9 +3,9 @@ author = "Nim Ethers Authors"
 description = "library for interacting with Ethereum"
 license = "MIT"
 
-requires "nim >= 1.6.0"
+requires "nim >= 1.6.16"
 requires "chronicles >= 0.10.3 & < 0.11.0"
-requires "chronos >= 3.0.0 & < 4.0.0"
+requires "chronos#head" # FIXME: change to >= 4.0.0 when chronos 4 is released
```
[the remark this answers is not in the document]
gmega: Yeah... I think both are bad TBH. :smile: The tradeoff I see is:

* using a specific commit buys you stability for the nim-ethers build in the short term, but will require you to constantly bump this to keep up with the latest V4 commits;
* using head means every time someone pushes you'll be testing against the latest pre-release of V4, and you may be able to catch issues earlier. Can be annoying to someone who just want to change unrelated stuff and all of a sudden finds themselves with a Chronos error unrelated to their change, though.

Either one sort of sucks IMHO, so I'd be fine with either.

emizzle:
Are we ok to replace this with https://github.com/codex-storage/nim-ethers/pull/64?

PR #63, opened 2024-02-05
Fix decoding of getter functions

markspanbroek:
Getter functions that are generated by the solidity compiler do not wrap their return value in a tuple like other functions do.

Depends on #61


emizzle:
https://github.com/codex-storage/nim-ethers/pull/64 is up for the json-rpc and chronos upgrades 👍

emizzle (approved):
Really awesome mark! Sorry for the rebase that will need to happen after #64 😅

markspanbroek:
> Sorry for the rebase that will need to happen after #64 😅

No problem, thanks for handling the chronos and jsonrpc updates! Rebasing is very simple in comparison :smile:

PR #64, opened 2024-02-08
Upgrade to `nim-json-rpc` v0.4.2 and chronos v4

emizzle:
`nim-json-rpc#head`, in particular has seen some large breaking changes:
- removal of `setMethodHandler` which is how ethers was processing subscription updates
- `call` returns `JsonString` instead of `JsonNode`
- json marshalling uses a custom flavour of `nim-json-serialization` exclusively instead of `%` and `fromJson`

This PR addresses those breaking changes and upgrades to chronos v4, which supports proper async exception tracking. Ethers now has proper exception tracking support.

Adding `{.push raises:[].}` as part of the exception tracking support causes `Tables` with `JsonNode` keys to raise an `Exception` (due to a call to `JsonNode.hash`). The underlying error was a missing forward declaration for `hash` of a particular type. The workaround for this was to create a shim that adds the missing forward declaration and also adds `{.raises: [].}`.

`nim-json-rpc` serialization replaced the `jsonmarshal` module with forced serialization via `nim-json-serialization`. As `nim-json-serialization` is not a preferred serializer, json serialization was replaced with [`nim-serde`](https://github.com/codex-storage/nim-serde).


ethers/providers/jsonrpc.nim
```diff
@@ -1,18 +1,17 @@
...
 import pkg/json_rpc/errors
+import pkg/serde
 import ../basics
 import ../provider
 import ../signer
 import ./jsonrpc/rpccalls
 import ./jsonrpc/conversions
 import ./jsonrpc/subscriptions
 
-export json
```
markspanbroek, line 15: I'm so glad that we don't need to do this anymore :+1:

ethers/signer.nim
```diff
@@ -1,15 +1,16 @@
...
+{.push raises: [].}
+
 type
   Signer* = ref object of RootObj
     lastSeenNonce: ?UInt256
     populateLock: AsyncLock
-
-type
-  SignerError* = object of EthersError
+  SignerError* = object of ProviderError
```
markspanbroek, line 13: I'd rather not have `SignerError` be a `ProviderError`. Mostly because a signer doesn't have to be associated with a provider. For instance, a wallet can sign stuff without involving a provider.

testmodule/providers/jsonrpc/testConversions.nim
```diff
@@ -17,14 +20,16 @@ suite "JSON Conversions":
       "timestamp":"0x6285c293"
     }
 
-    var blk = Block.fromJson(json)
-    check blk.number.isNone
+    without blk1 =? Block.fromJson(json):
+      fail
```
markspanbroek, line 24: I'd do this:

```suggestion
    without blk1 =? Block.fromJson(json):
      fail()
```
Or even this:
```suggestion
    var blk = !Block.fromJson(json)
```

I don't like using `!`, except in tests like this.

markspanbroek (approved):
Very nice, I like these changes a lot. They make exception handling a lot better, and the improved json serialization makes everything simpler.

ethers/providers/jsonrpc/subscriptions.nim
```diff
@@ -56,18 +94,18 @@ proc new*(_: type JsonRpcSubscriptions,
           client: RpcWebSocketClient): JsonRpcSubscriptions =
   let subscriptions = WebSocketSubscriptions(client: client)
   proc subscriptionHandler(arguments: JsonNode) {.raises:[].} =
-    if id =? arguments["subscription"].catch and
+    if id =? arguments{"subscription"}.catch and
```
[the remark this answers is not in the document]
emizzle: If it is `nil`, then calling `getCallback` with `nil` could SIGSEGV. Checking for `and id != nil` raises `KeyError`. So in an attempt to not raise errors in `new`, I've changed added the nil check to `getCallback` (which raises `KeyError`), but now we can catch that inside of `new`.

ethers/providers/jsonrpc/subscriptions.nim
```diff
@@ -1,19 +1,57 @@
...
+    methodHandlers: Table[string, MethodHandler]
+  MethodHandler* = proc (j: JsonNode) {.gcsafe, raises: [].}
   SubscriptionCallback = proc(id, arguments: JsonNode) {.gcsafe, raises:[].}
 
+{.push raises:[].}
+
+template `or`(a: JsonNode, b: typed): JsonNode =
+  if a.isNil: b else: a
+
+func init*(subscriptions: JsonRpcSubscriptions) =
```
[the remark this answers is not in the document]
emizzle: The main purpose was to allow `new` to create an instance of JsonRpcSubscriptions without initializing services, such as parsing of every message, which may be unnecessary or which can lead to race conditions, when the caller simply needed an object instance.

I am not opposed to putting this in `new` if you feel strongly about it.

ethers/signer.nim
emizzle: I had another look at this, and separated `ProviderError` from `SignerError` so we can keep their functionality separate, as Ledger does. Some signer methods, eg `getChainId` now can raise both `ProviderError` and `SignerError`. `reverts` additionally checks for `SignerError` now as well 👍

markspanbroek (approved):

PR #70, opened 2024-03-18
Update dependencies

markspanbroek:
Updates hardhat and solidity

~~Removes `increaseAllowance` and `decreaseAllowance` from ERC20, because they are removed from openzeppelin's ERC20 implementation, and are replaced by [SafeERC20](https://docs.openzeppelin.com/contracts/5.x/api/token/erc20#SafeERC20)~~

Uses `personal_sign` instead of `eth_sign`, because ethers.js also uses `personal_sign`, and `eth_sign` is now broken in hardhat (arguments are reversed).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,7 +18,7 @@ jobs:
     - name: Install NodeJS
       uses: actions/setup-node@v2
       with:
-        node-version: '14'
+        node-version: '18'
     - name: Install test node
       working-directory: testnode
       run: npm install
--- a/ethers/erc20.nim
+++ b/ethers/erc20.nim
@@ -54,24 +54,6 @@ method approve*(token: Erc20Token,
                 amount: UInt256): ?TransactionResponse {.base, contract.}
   ## Sets `amount` as the allowance of `spender` over the caller's tokens.
 
-method increaseAllowance*(token: Erc20Token,
-                spender: Address,
-                addedValue: UInt256): ?TransactionResponse {.base, contract.}
-  ## Atomically increases the allowance granted to spender by the caller.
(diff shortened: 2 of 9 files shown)


AuHau:
> Removes increaseAllowance and decreaseAllowance from ERC20, because they are removed from openzeppelin's ERC20 implementation, and are replaced by [SafeERC20](https://docs.openzeppelin.com/contracts/5.x/api/token/erc20#SafeERC20)

I am not sure about this. First of all, we use these in `nim-codex`, so we will have to bring those back in nim-codex. Also from the perspective of the potential users of this. There are still many ERC20 tokens deployed out there that have these functions (and it was under big warning that the user should verify if their token support these, so they should double-check anyhow). 

I think I would suggest keeping these.

markspanbroek:
> I am not sure about this. First of all, we use these in `nim-codex`, so we will have to bring those back in nim-codex. 

We could add these functions to `nim-codex` instead of in this repo. However, as soon as we upgrade openzeppelin in `codex-contracts-eth`, we'll lose those functions on our `TestToken` anyway. So we'd better start thinking about how to use SafeERC20 instead :thinking: 

> Also from the perspective of the potential users of this. There are still many ERC20 tokens deployed out there that have these functions (and it was under big warning that the user should verify if their token support these, so they should double-check anyhow).
>
> I think I would suggest keeping these.

The main problem with keeping these, is that we cannot test them because they are not implemented in `ERC20.sol` anymore.

markspanbroek:
After discussion with @AuHau, we decided to keep `increaseAllowance` and `decreaseAllowance` by not upgrading openzeppelin for now. We can upgrade openzeppelin later, once we've removed our dependency on these functions in Codex: https://github.com/codex-storage/nim-codex/issues/598

PR #73, opened 2024-05-17
fix: deserialize BlockTag from empty string

emizzle:
Allows BlockTag to be deserialized from an empty string.


markspanbroek (approved):
Nice!

PR #74, opened 2024-05-17
Remove overloaded UInt256.fromJson

emizzle:
Rely instead on `UInt256.fromJson` from `nim-serde` to avoid overload symbol collisions. 

Bump nim-serde to version that contains [a workaround](https://github.com/codex-storage/nim-serde/pull/24) for the following:

When calling `createRpcSigs` from `nim-json-rpc`, the `readValue` mixin is called, which captures all `readValue` overloads in scope. The issue is that the `readValue` overload in ethers calls a generic `fromJson`, and only the `fromJson` generic overloads defined in ethers are in scope. However, implementing applications, eg `nim-codex` defined other `fromJson` generic overloads that are not in scope at the time of the `readValue` mixin call. The important part of this issue is that both `nim-codex` and `ethers` overload the same `fromJson` (`UInt256`). The result, is that when `nim-codex` calls `UInt256.fromJson`, it will not know about the overload in `nim-codex`, and instead always call the `ethers` overload, which is not correct.

This PR removes the ethers `UInt256.fromJson` overload as a partial solution to the issue. This overload is not necessary as the `UInt256.fromJson` achieves mostly the same result. The difference is that empty strings were deserialized into `0.u256`. This could possibly affect:
1. `BlockTag` from being deserialized when there's an empty string: fixed in https://github.com/codex-storage/nim-ethers/pull/73.
2. `blockNumber` (UInt256) from being deserialized when there's an empty string. It appears that deserializing a missing block number from a TransactionReceipt into 0 might have actually caused some issues when waiting on block confirmations, and this change *should* alleviate any of those issues.

 are not *all* yet in scope at the time of the `readValue` mixin call (some are, but others are defined in an implementing application, eg `nim-codex`). This basically forces `readValue` to use the `fromJson` from ethers which get muddled up in the `readValue` mixin scope (), which deserializes an empty string for `?UInt256` into `UInt256.none`. Previously, .


gmega (approved):
LGTM!

ethers.nimble
```diff
@@ -9,7 +9,7 @@ requires "chronos >= 4.0.0 & < 4.1.0"
 requires "contractabi >= 0.6.0 & < 0.7.0"
 requires "questionable >= 0.10.2 & < 0.11.0"
 requires "json_rpc >= 0.4.0 & < 0.5.0"
-requires "serde >= 0.1.1 & < 0.2.0"
+requires "serde#e889b76dda0636421484dcda1535aa0911345244"
```
markspanbroek, line 12: I don't want ethers to be pinned to a specific git hash of serde. Can't you make a release of serde instead?

markspanbroek (approved):
I like it. It removes code :grin:

Please address the comment, then I think it's good to go!

ethers.nimble
emizzle: Yes, good catch. This was meant to be changed when the serde PR was merged and a release cut. Updated.

PR #77, opened 2024-10-16
fix: pinning nim-eth dependency

AuHau:


emizzle (approved):

emizzle:
Pinning `nim-eth` is a temporary fix until proper versioning is implemented.

PR #81, opened 2024-10-23
fix: nonce too high

emizzle:
Concurrent asynchronous population of transactions cause issues with nonces not being in sync with the transaction count for an account on chain. This was being mitigated by tracking a "last seen" nonce and locking inside of `populateTransaction` so that the nonce could be populated in a concurrent fashion. However, if there was an async cancellation before the transaction was sent, then the nonce would become out of sync. One solution was to decrease the nonce if a cancellation occurred. The other solution, in this PR, is simply to lock the populate and sendTransaction calls together, so that there will not be concurrent nonce discrepancies. This removes the need for "lastSeenNonce" and is overall more simple.

This issue was causing issues in Codex's testnet where slots were not being filled due to incorrect nonces and pending transactions.

### DIFF
--- a/ethers/contract.nim
+++ b/ethers/contract.nim
@@ -127,10 +127,17 @@ proc send(contract: Contract,
           overrides = TransactionOverrides()):
          Future[?TransactionResponse] {.async.} =
   if signer =? contract.signer:
-    let transaction = createTransaction(contract, function, parameters, overrides)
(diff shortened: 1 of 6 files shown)


markspanbroek (approved):
Very nice fix, I like that you were able to remove some of the complexity around the nonce tracking!

AuHau (approved):
LGTM, nice!

PR #82, opened 2024-10-23
fix: modify unsubscribe cleanup routine

emizzle:
Ignore exceptions (other than CancelledError) if uninstallation of the filter fails. If it's the last step in the subscription cleanup, then filter changes for this filter will no longer be polled so if the filter continues to live on in geth for whatever reason, then it doesn't matter.


emizzle:
Closing in favour of #84

PR #84, opened 2024-10-24
fix: modify unsubscribe cleanup routine and tests

### LINKED ISSUE #79: eth_getFilterChanges occasionally returns JNull on Windows
On Windows when running tests, occasionally tests were failing with the following error:

```
`node.kind == JArray` : items() can not iterate a JsonNode of kind JNull"
```

the error is caused by `eth_getFilterChanges` occasionally returning `JNull` instead of `JArray` on Windows.

This is happening in [vendor/nim-ethers/ethers/providers/jsonrpc/subscriptions.nim](https://github.com/codex-storage/nim-ethers/blob/e15974eb1f0ef3b7ed50b6f6529244d6047229d0/ethers/providers/jsonrpc/subscriptions.nim#L148):

```nim
proc getChanges(id: JsonNode): Future[JsonNode] {.async.} =
  try:
    return await subscriptions.client.eth_getFilterChanges(id)
  except CatchableError:
    return newJArray()

proc poll(id: JsonNode) {.async.} =
  for change in await getChanges(id):
    if callback =? subscriptions.getCallback(id):
      callback(id, change)
```

The code is basically correct, as _should_ always return a `JArray`. A fix could e.g.:

```nim
proc poll(id: JsonNode) {.async.} =
  let changes = await getChanges(id)
  if changes.kind != JArray:
    return
  for change in changes:
    if callback =? subscriptions.getCallback(id):
      callback(id, change)
```

Alternatively, we could also modify the `getChanges` proc.

A similar issue were reported in the past for go-ethereum: https://github.com/ethereum/go-ethereum/issues/2746, but it should be resolved by now. Maybe somehow it propagated to HardHat and apparently on Windows only - I could not find any issue at Hardhat's GitHub.

emizzle:
Closes #79 

Modifying the cleanup:
- Ignore exceptions (other than CancelledError) if uninstallation of the filter fails. If it's the last step in the subscription cleanup, then filter changes for this filter will no longer be polled so if the filter continues to live on in geth for whatever reason, then it doesn't matter.

To fix the subscription filter tests:
- `CancelledError` is now caught inside of `getChanges`. This was causing conditions during `subscriptions.close`, where the `CancelledError` would get consumed by the `except CatchableError`, if there was an ongoing `poll` happening at the time of close.
- After creating a new filter inside of `getChanges`, the new filter is polled for changes before returning.
- `getChanges` also does not swallow `CatchableError` by returning an empty array, and instead re-raises the error if it is not `filter not found`.
- #79 was fixed by checking the return value of `eth_getFilterChanges`. If `JNull` was received, then an empty array is returned.
- The tests were simplified by accessing the private fields of `PollingSubscriptions`. That way, there wasn't a race condition for the `newFilterId` counter inside of the mock.
- The `MockRpcHttpServer` was simplified by keeping track of the active filters only, and invalidation simply removes the filter. The tests then only needed to rely on the fact that the filter id changed in the mapping.
- Because of the above changes, we no longer needed to sleep inside of the tests, so the sleeps were removed, and the polling interval could be changed to 1ms, which not only makes the tests faster, but would further highlight any race conditions if present.

Co-authored-by: Adam Uhlíř <adam@uhlir.dev>

### DIFF
--- a/ethers/providers/jsonrpc/subscriptions.nim
+++ b/ethers/providers/jsonrpc/subscriptions.nim
@@ -17,6 +17,7 @@ type
     methodHandlers: Table[string, MethodHandler]
   MethodHandler* = proc (j: JsonNode) {.gcsafe, raises: [].}
   SubscriptionCallback = proc(id, arguments: JsonNode) {.gcsafe, raises:[].}
+  SubscriptionError* = object of EthersError
 
 {.push raises:[].}
 
@@ -130,7 +131,7 @@ method unsubscribe*(subscriptions: WebSocketSubscriptions,
 # Polling
 
 type
-  PollingSubscriptions = ref object of JsonRpcSubscriptions
+  PollingSubscriptions* = ref object of JsonRpcSubscriptions
     polling: Future[void]
 
     # We need to keep around the filters that are used to create log filters on the RPC node
@@ -151,17 +152,24 @@ proc new*(_: type JsonRpcSubscriptions,
   proc getChanges(originalId: JsonNode): Future[JsonNode] {.async.} =
     try:
       let mappedId = subscriptions.subscriptionMapping[originalId]
-      echo "getting changes for originalId: ", $originalId, "; mappedId: ", $mappedId
-      return await subscriptions.client.eth_getFilterChanges(mappedId)
+      let changes = await subscriptions.client.eth_getFilterChanges(mappedId)
+      if changes.kind == JNull:
+        return newJArray()
+      elif changes.kind != JArray:
+        raise newException(SubscriptionError,
+          "HTTP polling: unexpected value returned from eth_getFilterChanges." &
+          " Expected: JArray, got: " & $changes.kind)
+      return changes
+    except CancelledError as e:
+      raise e
     except CatchableError as e:
       if "filter not found" in e.msg:
-        echo "filter not found - recreating; originalId: ", originalId
         let filter = subscriptions.filters[originalId]
         let newId = await subscriptions.client.eth_newFilter(filter)
-        echo "filter not found - newId: ", newId
         subscriptions.subscriptionMapping[originalId] = newId
-
-      return newJArray()
+        return await getChanges(originalId)
+      else:
+        raise e
 
   proc poll(id: JsonNode) {.async.} =
     for change in await getChanges(id):
@@ -171,9 +179,7 @@ proc new*(_: type JsonRpcSubscriptions,
   proc poll {.async.} =
     untilCancelled:
       for id in toSeq subscriptions.callbacks.keys:
-        echo "polling for key:", $id
         await poll(id)
-      echo "sleeping"
       await sleepAsync(pollingInterval)
 
   subscriptions.polling = poll()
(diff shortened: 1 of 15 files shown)


ethers/providers/jsonrpc/subscriptions.nim
```diff
@@ -151,14 +152,24 @@ proc new*(_: type JsonRpcSubscriptions,
...
+      let changes = await subscriptions.client.eth_getFilterChanges(mappedId)
+      if changes.kind == JNull:
+        return newJArray()
+      elif changes.kind != JArray:
+        raise newException(SubscriptionError,
+          "HTTP polling: unexpected value returned from eth_getFilterChanges." &
+          " Expected: JArray, got: " & $changes.kind)
+      return changes
+    except CancelledError as e:
+      raise e
```
AuHau, line 164: I would not raise inside polling loop, but we have planned the Result approach so I will pick that up based on this work.

ethers/providers/jsonrpc/subscriptions.nim
```diff
@@ -151,14 +152,24 @@ proc new*(_: type JsonRpcSubscriptions,
...
     except CatchableError as e:
       if "filter not found" in e.msg:
         let filter = subscriptions.filters[originalId]
         let newId = await subscriptions.client.eth_newFilter(filter)
         subscriptions.subscriptionMapping[originalId] = newId
```
AuHau, line 169: Btw. we are leaving a place for race condition here. While its probability is small, it can still happen. The `await` here yields the processing of this, so if meanwhile for example `unsubscribe` is called, then it will use the old ID because the new one was not yet processed. I will leave up to you if you want to mitigate that or will leave it as is.

testmodule/providers/jsonrpc/testJsonRpcSubscriptions.nim
```diff
@@ -122,35 +128,26 @@ suite "HTTP polling subscriptions - filter not found":
...
-    let jsonId = await subscriptions.subscribeLogs(filter, emptyHandler)
-    let id = string.fromJson(jsonId).tryGet
-    check mockServer.newFilterCounter == 1
+    check subscriptions.filters.len == 0
+    check subscriptions.subscriptionMapping.len == 0
+
+    let id = await subscriptions.subscribeLogs(filter, emptyHandler)
+
+    check subscriptions.filters[id] == filter
+    check subscriptions.subscriptionMapping[id] == id
```
AuHau, line 137: Maybe add check that `len == 1`?

AuHau (approved):
Few comments that I will leave up to you, but otherwise LGTM

testmodule/providers/jsonrpc/testJsonRpcSubscriptions.nim
emizzle: Updated.

ethers/providers/jsonrpc/subscriptions.nim
emizzle: To be addressed in https://github.com/codex-storage/nim-ethers/issues/86.

PR #85, opened 2024-10-24
fix: block filters can be also recreated

AuHau:
Adds also support for the recreation of blocks filters.

### DIFF
--- a/ethers/providers/jsonrpc/subscriptions.nim
+++ b/ethers/providers/jsonrpc/subscriptions.nim
@@ -140,7 +140,7 @@ type
     # We need to keep around the filters that are used to create log filters on the RPC node
     # as there might be a time when they need to be recreated as RPC node might prune/forget
     # about them
-    filters: Table[JsonNode, EventFilter]
+    logFilters: Table[JsonNode, EventFilter]
 
     # Used when filters are recreated to translate from the id that user
     # originally got returned to new filter id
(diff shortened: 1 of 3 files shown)


ethers/providers/jsonrpc/subscriptions.nim
```diff
@@ -164,8 +164,14 @@ proc new*(_: type JsonRpcSubscriptions,
...
       if "filter not found" in e.msg:
-        let filter = subscriptions.filters[originalId]
-        let newId = await subscriptions.client.eth_newFilter(filter)
+        var newId: JsonNode
+        # If there exists filter for given ID, then the filter was a log filter
+        # otherwise it was a block filter
+        if subscriptions.filters.hasKey(originalId):
+          let filter = subscriptions.filters[originalId]
+          newId = await subscriptions.client.eth_newFilter(filter)
+        else:
```
emizzle, line 173: This seems loose... maybe we could store an object in `subscriptionMapping` that has an `id` and `kind`, and then here we can check the `kind` before deciding what to do...

AuHau: I was also thinking a bit about this "looseness", but I did not really find any reason why it should lead to bug. 

If there is a callback (and hence, that is why this function is called), then the callback had to be created either through `subscribeLogs,` and hence it will have an entry in `filters` or through `suscribeBlocks` and then it will not have an entry in `filters`. IMHO this is good enough reasoning. If you think there will "magically disappear filters" then we have a problem and a bug somewhere anyway and will have to track that case down...

Soooo I would leave it as is.

emizzle: > If you think there will "magically disappear filters" then we have a problem and a bug somewhere anyway and will have to track that case down...

It's more about if `filters` functionality is changed at some point in the future, then we may introduce a bug that isn't easy to spot/debug.

It's also not the most easy to reason about... As I was reading the change, i thought "why wouldn't `filters` also contain a *block* filter?" Then I had to look at how creating a block filter doesn't add to `filters`. 

It seems a bit fragile honestly.

AuHau: Alright, so what about renaming `filters` to `logFilters` in order to better communicate it? Because that is what that really is. In the context of this Table, there are no "block filters", because block filters are only subscriptions for new blocks, not for "notify me when some specific blocks arrive" and hence it does not really make sense to store them in that Table...

emizzle: Yep, that works 👍

emizzle: > there are no "block filters"

Might be worth updating the comment?

emizzle (approved):
The `filters` > `logFilters` comment isn't a blocker, so approving now 👍

PR #87, opened 2024-10-25
fix: `items(JsonNode)` symbol not found

emizzle:
- Fixes `items(JsonNode)` symbol not found when calling `JsonRpcProvider.new` from a consumer. Exporting `serde` from `subscriptions` and then exporting `subscriptions` from `jsonrpc` allowed the symbols to be resolved in the consumer.

- Replaces all instances of `std/json` with `pkg/serde`

### DIFF
--- a/ethers/contract.nim
+++ b/ethers/contract.nim
@@ -1,4 +1,4 @@
-import std/json
+import pkg/serde
 import std/macros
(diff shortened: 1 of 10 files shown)


AuHau (approved):
Nice!

PR #88, opened 2024-10-25
fix: block filters can be also recreated

AuHau:
Adds Result to the Subscription's handlers that will return errors when any error in the underlaying subscription mechanism (for example HTTP polling) surface.

### DIFF
--- a/ethers/contract.nim
+++ b/ethers/contract.nim
@@ -3,6 +3,7 @@ import std/macros
 import std/sequtils
 import pkg/chronicles
 import pkg/chronos
+import pkg/questionable
 import pkg/contractabi
 import ./basics
 import ./provider
@@ -35,11 +36,10 @@ type
     gasLimit*: ?UInt256
   CallOverrides* = ref object of TransactionOverrides
     blockTag*: ?BlockTag
-  ContractError* = object of EthersError
   Confirmable* = object
     response*: ?TransactionResponse
     convert*: ConvertCustomErrors
-  EventHandler*[E: Event] = proc(event: E) {.gcsafe, raises:[].}
+  EventHandler*[E: Event] = proc(event: ?!E) {.gcsafe, raises:[].}
 
 func new*(ContractType: type Contract,
           address: Address,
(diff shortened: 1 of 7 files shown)


ethers/contract.nim
```diff
@@ -292,9 +292,13 @@ proc subscribe*[E: Event](contract: Contract,
...
 
-  proc logHandler(log: Log) {.raises: [].} =
+  proc logHandler(logResult: ?!Log) {.raises: [].} =
+    without log ?= logResult, err:
+      handler(failure(e))
+      return
+
     if event =? E.decode(log.data, log.topics):
-      handler(event)
+      handler(success(event))
```
emizzle, line 301: Wouldn't you just want to pass in the result of `E.decode`? It returns `?!E` already, and that way decoding errors won't get swallowed.

ethers/provider.nim
```diff
@@ -227,21 +226,28 @@ proc confirm*(
...
     blockNumber = number
     blockEvent.fire()
 
-  proc onBlock(blck: Block) =
+  proc onBlock(blckResult: SubscriptionResult[Block]) =
+    if blckResult.isErr:
+      blockSubscriptionError = blckResult.error()
+      blockEvent.fire()
+
+    let blck = blckResult.value
```
emizzle, line 249: Needs a `return` in the `if` block. Also, `blockSubscriptionError = nil` should be set somewhere to "reset" the error?

Can also do:
```nim
blockSubscriptionError = nil

without blck =? blckResult, error:
  blockSubscriptionError = error
  blockEvent.fire()
```

ethers/errors.nim
```diff
@@ -1,6 +1,16 @@
 import ./basics
 
-type SolidityError* = object of EthersError
+type
+  SolidityError* = object of EthersError
+  ContractError* = object of EthersError
+  SignerError* = object of EthersError
+  SubscriptionError* = object of EthersError
+  SubscriptionResult*[E] = Result[E, ref SubscriptionError]
```
emizzle, line 8: `SubscriptionResult` definitely works... but sometimes it can be a more of a pain than it's worth, mostly when needing to convert between different `Result` types when only the error type is different (eg using some flavour of `mapErr`). You can still return specific error types when it needs inspection at the callsite (since they inherit from CatchableError). See the `reservations` module for examples on how to do this, eg https://github.com/codex-storage/nim-codex/blob/ebee85b1aae05e50839a8f7e911ea888cff209bc/codex/sales/reservations.nim#L273 -- notice the return type of `updateAvailability` is `?!void`, but here we're returning a specific error type). This is the pattern I'd suggest: return/parameter types should `?!T`, then return specific error types where needed.

PR #91, opened 2024-10-31
feat: subscriptions get passed result questionable

AuHau:
Introduces breaking change to API, where subscription handlers now return Result type. Handlers, similarly to now will be called with events wrapped in the Result when new events are detected, but also handlers will be called when some error in the underlying event detection mechanism is detected.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,6 +17,8 @@ jobs:
     steps:
     - name: Checkout
       uses: actions/checkout@v4
+      with:
+        ref: ${{ github.event.pull_request.head.sha }}
 
     - name: Install Nim
       uses: iffy/install-nim@v4
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,6 +1,11 @@
 name: CI
 
-on: [push, pull_request, workflow_dispatch]
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+  workflow_dispatch:
 
 jobs:
   test:
--- a/ethers/errors.nim
+++ b/ethers/errors.nim
@@ -1,7 +1,23 @@
 import ./basics
 
-type SolidityError* = object of EthersError
+type
+  SolidityError* = object of EthersError
+  ContractError* = object of EthersError
+  SignerError* = object of EthersError
+  SubscriptionError* = object of EthersError
+  ProviderError* = object of EthersError
+    data*: ?seq[byte]
+
+template raiseSignerError*(message: string, parent: ref ProviderError = nil) =
+  raise newException(SignerError, message, parent)
 
 {.push raises:[].}
+
+proc toErr*[E1: ref CatchableError, E2: EthersError](
+  e1: E1,
+  _: type E2,
+  msg: string = e1.msg): ref E2 =
+
+  return newException(E2, msg, e1)
 
 template errors*(types) {.pragma.}
--- a/ethers/errors.nim
+++ b/ethers/errors.nim
@@ -8,9 +8,6 @@ type
   ProviderError* = object of EthersError
     data*: ?seq[byte]
 
-template raiseSignerError*(message: string, parent: ref ProviderError = nil) =
-  raise newException(SignerError, message, parent)
-
 {.push raises:[].}
 
 proc toErr*[E1: ref CatchableError, E2: EthersError](
--- a/ethers/providers/jsonrpc/subscriptions.nim
+++ b/ethers/providers/jsonrpc/subscriptions.nim
@@ -1,9 +1,12 @@
 import std/tables
 import std/sequtils
+import std/strutils
 import pkg/chronos
+import pkg/questionable
 import pkg/json_rpc/rpcclient
 import pkg/serde
 import ../../basics
+import ../../errors
 import ../../provider
 include ../../nimshims/hashes
 import ./rpccalls
@@ -17,8 +20,7 @@ type
     callbacks: Table[JsonNode, SubscriptionCallback]
     methodHandlers: Table[string, MethodHandler]
   MethodHandler* = proc (j: JsonNode) {.gcsafe, raises: [].}
-  SubscriptionCallback = proc(id, arguments: JsonNode) {.gcsafe, raises:[].}
-  SubscriptionError* = object of EthersError
+  SubscriptionCallback = proc(id: JsonNode, arguments: ?!JsonNode) {.gcsafe, raises:[].}
 
 {.push raises:[].}
 
@@ -54,7 +56,7 @@ proc setMethodHandler(
 method subscribeBlocks*(subscriptions: JsonRpcSubscriptions,
                         onBlock: BlockHandler):
                        Future[JsonNode]
-                       {.async, base.} =
+                       {.async, base, raises: [CancelledError].} =
   raiseAssert "not implemented"
 
 method subscribeLogs*(subscriptions: JsonRpcSubscriptions,
@@ -75,14 +77,16 @@ method close*(subscriptions: JsonRpcSubscriptions) {.async, base.} =
     await subscriptions.unsubscribe(id)
 
 proc getCallback(subscriptions: JsonRpcSubscriptions,
-                 id: JsonNode): ?SubscriptionCallback =
+                 id: JsonNode): ?SubscriptionCallback  {. raises:[].} =
   try:
     if not id.isNil and id in subscriptions.callbacks:
-      subscriptions.callbacks[id].some
+        try:
+          return subscriptions.callbacks[id].some
+        except: discard
     else:
-      SubscriptionCallback.none
+      return SubscriptionCallback.none
   except KeyError:
-    SubscriptionCallback.none
+    return SubscriptionCallback.none
 
 # Web sockets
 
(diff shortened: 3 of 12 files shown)


ethers/providers/jsonrpc/subscriptions.nim
```diff
@@ -77,14 +78,16 @@ method close*(subscriptions: JsonRpcSubscriptions) {.async, base.} =
...
-      subscriptions.callbacks[id].some
+        try:
+          return subscriptions.callbacks[id].some
+        except: discard
     else:
-      SubscriptionCallback.none
+      return SubscriptionCallback.none
   except KeyError:
-    SubscriptionCallback.none
+    return SubscriptionCallback.none
```
emizzle, line 90: spacing lines 84-86

What exception was leaking from `subscriptions.callbacks[id].some`? 

By default, SubscriptionCallback.none is returned, so we don't really need line 87-88.

The catching of `KeyError` is not needed with the try/except since it's caught on 86. If you're going to use try/except discard, then this whole thing can be a bit nicer:

```nim 
if not id.isNil and id in subscriptions.callbacks:
  try:
    return subscriptions.callbacks[id].some
  except: discard
```

ethers/providers/jsonrpc/subscriptions.nim
```diff
@@ -179,12 +192,19 @@ proc new*(_: type JsonRpcSubscriptions,
         subscriptions.subscriptionMapping[originalId] = newId
         return await getChanges(originalId)
       else:
-        raise e
+        raise newException(SubscriptionError, "HTTP polling: There was an exception while getting subscription changes: " & e.msg, e)
+
+  proc poll(id: JsonNode) {.async: (raises: [CancelledError, SubscriptionError]).} =
+    without callback =? subscriptions.getCallback(id):
+        return
```
emizzle, line 199: spacing

ethers/providers/jsonrpc/subscriptions.nim
```diff
@@ -202,16 +222,23 @@ method close*(subscriptions: PollingSubscriptions) {.async.} =
...
     try:
       if blck =? (await subscriptions.client.eth_getBlockByHash(hash, false)):
-        onBlock(blck)
-    except CatchableError:
+        onBlock(success(blck))
+    except CancelledError as e:
       discard
+    except CatchableError as e:
+      let wrappedErr = newException(SubscriptionError, "HTTP polling: There was an exception while getting subscription's block: " & e.msg, e)
+      onBlock(failure(Block, wrappedErr))
```
emizzle, line 235: 🟡  can also do:
```suggestion
      let err = e.toErr(SubscriptionError,
        "HTTP polling: There was an exception while getting subscription's block: " & e.msg)
      onBlock(failure(Block, err))
```

ethers/providers/jsonrpc/subscriptions.nim
```diff
@@ -202,16 +222,23 @@ method close*(subscriptions: PollingSubscriptions) {.async.} =
...
+
+  proc callback(id: JsonNode, changeResult: ?!JsonNode) {.raises:[].} =
+    without change =? changeResult, error:
+        onBlock(failure(Block, error.toErr(SubscriptionError)))
+        return
```
emizzle, line 240: spacing

ethers/providers/jsonrpc/subscriptions.nim
```diff
@@ -226,9 +253,13 @@ method subscribeLogs(subscriptions: PollingSubscriptions,
                     Future[JsonNode]
                     {.async.} =
 
-  proc callback(id, change: JsonNode) =
+  proc callback(id: JsonNode, changeResult: ?!JsonNode) =
+    without change =? changeResult, error:
+        onLog(failure(Log, error.toErr(SubscriptionError)))
+        return
```
emizzle, line 259: spacing, fml

.github/workflows/ci.yml
```diff
@@ -1,6 +1,11 @@
 name: CI
 
-on: [push, pull_request, workflow_dispatch]
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+  workflow_dispatch:
```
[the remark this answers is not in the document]
AuHau: Because in the way we have it till now the CI gets "double triggered" because of PR and "push". See: https://discord.com/channels/895609329053474826/963349623714504784/1301528920893558855

ethers/providers/jsonrpc/subscriptions.nim
```diff
@@ -179,12 +192,19 @@ proc new*(_: type JsonRpcSubscriptions,
         subscriptions.subscriptionMapping[originalId] = newId
         return await getChanges(originalId)
       else:
-        raise e
+        raise newException(SubscriptionError, "HTTP polling: There was an exception while getting subscription changes: " & e.msg, e)
```
emizzle, line 195: 🟡  I'm wondering now if maybe `poll` and `getChanges` should return Results instead of raising or re-raising to make things more readable. They should still be able to raise `CancelledError` though. Maybe not, just a thought I had while thinking about adding in the ability to get logs for events that happen between losing a subscription and re-subscribing.

AuHau: Not sure what you mean. Yes, we throw an exception here, but this is not propagated above `poll` call site, where it is caught and then returned as a failed result into the callback...

https://github.com/codex-storage/nim-ethers/pull/91/files/3c64903822b15c7de2d6620a3ffbea0e879fc772#diff-597f9431d9147b024d3410322caa9432b95c8543f1c21c1c6ed784c9704fa943R203

Does that address what are you suggesting here?

ethers/errors.nim
```diff
@@ -1,7 +1,23 @@
...
+type
+  SolidityError* = object of EthersError
+  ContractError* = object of EthersError
+  SignerError* = object of EthersError
+  SubscriptionError* = object of EthersError
+  ProviderError* = object of EthersError
+    data*: ?seq[byte]
+
+template raiseSignerError*(message: string, parent: ref ProviderError = nil) =
+  raise newException(SignerError, message, parent)
```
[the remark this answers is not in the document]
AuHau: You are right that it is not really "global". But my experience is that it is really a pain in the ass if you have scattered all the error types and related utilities (like for example, this template) in several "error modules" and then you always have to figure out which you need to import when you need certain error type. I would agree about this in context of bigger projects like nim-codex where there might be too many error types, but on such a small library like this IMHO, it has the benefit to have it "centralized" in one `errors.nim` in root... And that is what I was trying to gear towards with these changes... WDYT?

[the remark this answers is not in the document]
emizzle: If there are module-level errors in place, then importing the module to use it also imports the errors associated with that module. Typically, those module-level errors will also inherit from library-level errors, so you get those for free as well (as long they're exported from the module). I don't really mind, but my preference is to have module-level errors as it promotes module-level exception handling.

ethers/providers/jsonrpc/subscriptions.nim
emizzle: I was referring to having `getChanges` and `poll` return Results instead of bubbling exceptions which ultimately will get passed to callback:
```diff
- proc getChanges(id: JsonNode): Future[JsonNode] {.async: (raises: [CancelledError]).} =
+ proc getChanges(id: JsonNode): Future[?!JsonNode] {.async: (raises: [CancelledError]).} =
```
and
```diff
- proc poll(id: JsonNode) {.async: (raises: [CancelledError]).} =
+ proc poll(id: JsonNode): Future[?!void] {.async: (raises: [CancelledError]).} =
```

emizzle: There are two `poll` functions, I was referring to the one that does not call the callbacks.

ethers/providers/jsonrpc/subscriptions.nim
```diff
@@ -75,14 +77,14 @@ method close*(subscriptions: JsonRpcSubscriptions) {.async, base.} =
...
     if not id.isNil and id in subscriptions.callbacks:
-      subscriptions.callbacks[id].some
-    else:
-      SubscriptionCallback.none
+      try:
+        return subscriptions.callbacks[id].some
+      except: discard
   except KeyError:
-    SubscriptionCallback.none
+    return SubscriptionCallback.none
```
emizzle, line 87: `KeyError` will not leak from this, so no need to catch it. The entire block can be simplified to:
```nim
if not id.isNil and id in subscriptions.callbacks:
  try:
    return subscriptions.callbacks[id].some
  except: discard
```
OR
```nim
try:
  if not id.isNil and id in subscriptions.callbacks:
    return subscriptions.callbacks[id].some
except: discard
```

ethers/providers/jsonrpc/subscriptions.nim
```diff
@@ -116,9 +123,14 @@ method subscribeLogs(subscriptions: WebSocketSubscriptions,
...
                     {.async.} =
-  proc callback(id, arguments: JsonNode) =
+  proc callback(id: JsonNode, argumentsResult: ?!JsonNode) =
+    without arguments =? argumentsResult, error:
+      onLog(failure(Log, error.toErr(SubscriptionError)))
+      return
+
     if log =? Log.fromJson(arguments{"result"}):
-      onLog(log)
+      onLog(success(log))
```
emizzle, line 132: Should we also propagate deserialisation errors here?
```nim
let res = Log.fromJson(arguments{"result"}).mapFailure(SubscriptionError)
onLog(res)
```

ethers/providers/jsonrpc/subscriptions.nim
emizzle: Ah I see, looking at the code again, ignore that last comment 👍

ethers/providers/jsonrpc/subscriptions.nim
```diff
@@ -237,9 +269,13 @@ method subscribeLogs(subscriptions: PollingSubscriptions,
...
 
-  proc callback(id, change: JsonNode) =
+  proc callback(id: JsonNode, changeResult: ?!JsonNode) =
+    without change =? changeResult, error:
+      onLog(failure(Log, error.toErr(SubscriptionError)))
+      return
+
     if log =? Log.fromJson(change):
-      onLog(log)
+      onLog(success(log))
```
emizzle, line 278: Same thing regarding propagating deserialisation errors?

ethers/providers/jsonrpc/subscriptions.nim
```diff
@@ -213,16 +238,23 @@ method close*(subscriptions: PollingSubscriptions) {.async.} =
...
+                      {.async, raises:[CancelledError].} =
 
-  proc getBlock(hash: BlockHash) {.async.} =
+  proc getBlock(hash: BlockHash) {.async: (raises:[]).} =
     try:
       if blck =? (await subscriptions.client.eth_getBlockByHash(hash, false)):
-        onBlock(blck)
-    except CatchableError:
+        onBlock(success(blck))
+    except CancelledError as e:
```
emizzle, line 247:
```suggestion
    except CancelledError:
```

emizzle (commented):
I hate to ask, but do you think we could create a test for this? Maybe use the `MockRpcServer` to emulate a failure during an RPC call, which then gets propagated to the handler...

PR #94, opened 2024-11-11
Fix http subscription polling

markspanbroek:
Solves several issues with polling.

Includes 2 fixes in http subscription polling:
- resubscribe when a JSON RPC error is encountered, instead of looking for the string "filter not found" in the error message, because different nodes can have different error messages
- ensure that the polling loop cannot crash, by explicitly annotating the function signatures with allowed errors, and by using `asyncSpawn`

Includes 2 fixes in transaction confirmation:
- no longer allow `.confirm(0)`, because it doesn't wait for any blocks to be mined, not even the block that includes the transaction
- no longer rely on the block number that is reported in the newBlock filter callback, because hardhat is known to report pending blocks there

### DIFF
--- a/ethers/providers/jsonrpc/subscriptions.nim
+++ b/ethers/providers/jsonrpc/subscriptions.nim
@@ -1,6 +1,5 @@
 import std/tables
 import std/sequtils
-import std/strutils
 import pkg/chronos
 import pkg/json_rpc/rpcclient
 import pkg/serde
@@ -9,7 +8,6 @@ import ../../provider
 include ../../nimshims/hashes
 import ./rpccalls
 import ./conversions
-import ./looping
 
 export serde
 
--- a/ethers/providers/jsonrpc/subscriptions.nim
+++ b/ethers/providers/jsonrpc/subscriptions.nim
@@ -152,47 +152,62 @@ proc new*(_: type JsonRpcSubscriptions,
 
   let subscriptions = PollingSubscriptions(client: client)
 
-  proc getChanges(originalId: JsonNode): Future[JsonNode] {.async.} =
+  proc resubscribe(id: JsonNode) {.async: (raises: [CancelledError]).} =
     try:
-      let mappedId = subscriptions.subscriptionMapping[originalId]
-      let changes = await subscriptions.client.eth_getFilterChanges(mappedId)
-      if changes.kind == JNull:
-        return newJArray()
-      elif changes.kind != JArray:
-        raise newException(SubscriptionError,
-          "HTTP polling: unexpected value returned from eth_getFilterChanges." &
-          " Expected: JArray, got: " & $changes.kind)
-      return changes
-    except CancelledError as e:
-      raise e
-    except CatchableError as e:
-      if "filter not found" in e.msg:
-        var newId: JsonNode
-        # Log filters are stored in logFilters, block filters are not persisted
-        # there is they do not need any specific data for their recreation.
-        # We use this to determine if the filter was log or block filter here.
-        if subscriptions.logFilters.hasKey(originalId):
-          let filter = subscriptions.logFilters[originalId]
-          newId = await subscriptions.client.eth_newFilter(filter)
-        else:
-          newId = await subscriptions.client.eth_newBlockFilter()
-        subscriptions.subscriptionMapping[originalId] = newId
-        return await getChanges(originalId)
+      var newId: JsonNode
+      # Log filters are stored in logFilters, block filters are not persisted
+      # there is they do not need any specific data for their recreation.
+      # We use this to determine if the filter was log or block filter here.
+      if subscriptions.logFilters.hasKey(id):
+        let filter = subscriptions.logFilters[id]
+        newId = await subscriptions.client.eth_newFilter(filter)
       else:
-        raise e
+        newId = await subscriptions.client.eth_newBlockFilter()
+      subscriptions.subscriptionMapping[id] = newId
+    except CancelledError as error:
+      raise error
+    except CatchableError:
+      # there's nothing further we can do here
+      discard
 
-  proc poll(id: JsonNode) {.async.} =
+  proc getChanges(id: JsonNode): Future[JsonNode] {.async: (raises: [CancelledError]).} =
+    try:
+      let mappedId = subscriptions.subscriptionMapping[id]
+      let changes = await subscriptions.client.eth_getFilterChanges(mappedId)
+      if changes.kind == JArray:
+        return changes
+    except KeyError as error:
+      raiseAssert "subscription mapping invalid: " & error.msg
+    except JsonRpcError:
+      await resubscribe(id)
+      # TODO: we could still miss some events between losing the subscription
+      # and resubscribing. We should probably adopt a strategy like ethers.js,
+      # whereby we keep track of the latest block number that we've seen
+      # filter changes for:
+      # https://github.com/ethers-io/ethers.js/blob/f97b92bbb1bde22fcc44100af78d7f31602863ab/packages/providers/src.ts/base-provider.ts#L977
+    except CancelledError as error:
+      raise error
+    except CatchableError:
+      # there's nothing we can do here
+      discard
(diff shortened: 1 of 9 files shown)


markspanbroek:
> [We should probably not allow any exceptions to escape here since this function is async spawned, which will turn the exceptions into defects.](https://github.com/codex-storage/nim-ethers/commit/9cfdd50779e3c8129b10d85bfe28fc8d3eaeb6d2#diff-aeaf8c0718afc0e1a762acdd87776f9b34bec54dc69ee8da3dbc59894d55e461R243)

Good catch! Thanks @emizzle, this is fixed now.

> [If this is happening, getting the current block may be the same as the old block, so:](https://github.com/codex-storage/nim-ethers/commit/9cfdd50779e3c8129b10d85bfe28fc8d3eaeb6d2#diff-aeaf8c0718afc0e1a762acdd87776f9b34bec54dc69ee8da3dbc59894d55e461R250)
>
>    1. Maybe a comparison check is needed before the async event is fired
>    2. This might be a source of timing issues in the tests

  1. If we get pending blocks, then the number is going to be larger than the block number that we've already seen, so I don't think that the comparison helps.
  2.  Luckily the OnChainClock already has similar logic, so it should not be impacted.

testmodule/providers/jsonrpc/testJsonRpcSubscriptions.nim
```diff
@@ -174,3 +178,16 @@ suite "HTTP polling subscriptions - filter not found":
...
 
     check not subscriptions.subscriptionMapping.hasKey id
+
+  test "polling continues with new filter after temporary error":
+    let filter = EventFilter(address: Address.example, topics: @[array[32, byte].example])
+    let emptyHandler = proc(log: Log) = discard
+
+    let id = await subscriptions.subscribeLogs(filter, emptyHandler)
+
+    await stopServer()
```
emizzle, line 188: I tried this as well and the errors I received were not the same errors that were being swallowed in the testnet. I was only able to emulate the exact error by disconnecting the client without stopping it (by connecting it to a closed port).

ethers/providers/jsonrpc/subscriptions.nim
```diff
@@ -152,47 +150,60 @@ proc new*(_: type JsonRpcSubscriptions,
...
 
-  proc poll(id: JsonNode) {.async.} =
+  proc getChanges(id: JsonNode): Future[JsonNode] {.async: (raises: [CancelledError]).} =
+    if mappedId =? subscriptions.subscriptionMapping.?[id]:
+      try:
+        let changes = await subscriptions.client.eth_getFilterChanges(mappedId)
+        if changes.kind == JArray:
+          return changes
+      except JsonRpcError:
+        await resubscribe(id)
```
emizzle, line 178: My only reservation here is that we may be resubscribing in too many situations. In the case of a disconnection, we don't really need to resubscribe as the subscription should still be on the server. Imagine there's a disconnection for 1 minute. At four second polling intervals, that'll be 15 new subscriptions. It may contribute to bloated change sets...

The main idea before was to keep the polling loop from silently crashing, by restarting the polling loop when it errored. However, you have covered this by asyncSpawning the polling loop and ensuring there are no exceptions leaked from it.

emizzle (approved):
This is looking much better, thanks for doing this Mark.

My preference would be to see a test that covers the situation that brought this about in the testnet originally, where there is a disconnection from the RPC server end. However, it's a preference so will approve now.

emizzle:
Sorry, reviewing on the phone is difficult. The block number comparison was already included but I just didn't see it. 👍

testmodule/providers/jsonrpc/testJsonRpcSubscriptions.nim
markspanbroek: I just tried out connecting to a closed port, like you did [here](https://github.com/codex-storage/nim-ethers/commit/39d03f642cf6fa6e47f21c6aa3d70a728af68823#diff-10d724a619389f6f12a5d2743337651e9bd50084f7b9e7c084d4472f76fba6a1R141), but it gave me the exact same JsonRpcError "Failed to send POST Request with JSON-RPC: Could not connect to remote host" :shrug:

ethers/providers/jsonrpc/subscriptions.nim
markspanbroek: > Imagine there's a disconnection for 1 minute. At four second polling intervals, that'll be 15 new subscriptions.

Not if we're disconnected, then the re-subscriptions will also fail.

But I agree, this could be more elegant. Ideally I'd like to move to how it's done in ethers.js. If I understand it correctly, they only use a single filter to watch for new blocks, and they parse the block logs to see if there are any subscriptions in memory that need to be triggered.

emizzle: `JsonRpcError` is a base-level error, of which all other types of errors in nim-json-rpc are derived. If we re-subscribe on all errors of this type, we may be doing ourselves a disservice and re-subscribing unnecessarily. For example, the error types `ErrorResponse`, `InvalidRequest`, `ApplicationError`, etc are all derived from `JsonRpcError`. This means that we may have encountered an error that does not indicate the filters were discarded or not found server-side. In my opinion, we don't necessarily NEED to resubscribe unless we encountered a `filter not found` error, (guess: would be a type of `ErrorResponse`). So, I think maybe we should re-introduce this condition, eg:

```nim
proc getChanges() =
  try:
    getFilterChanges
  except ErrorResponse as e:
      await resubscribe(id)
  except CancelledError as e:
    raise e
  except CatchableError as e:
    discard
```
This would still require the additional work to ensure we don't miss events between losing the subscription and resubscription.

testmodule/providers/jsonrpc/testJsonRpcSubscriptions.nim
emizzle: My apologies, I remember now why this didn't work for me. I was testing that the subscription would survive across temporary disconnections, and because restarting the mock server required an instantiation of `MockRpcHttpServer.new`, the subscriptions were reset. Thus it was not sufficient for my liking, and why I resorted to connecting to an invalid port. This allowed me to keep the existing `MockRpcHttpServer` instance while restarting the failed polling loop on the client side.

ethers/providers/jsonrpc/subscriptions.nim
emizzle: I checked, and the error type in the RPC client would likely be a `JsonRpcError` instead of `ErrorResponse`, so the above block would be:
```nim
proc getChanges() =
  try:
    getFilterChanges
  except JsonRpcError as e:
      await resubscribe(id)
  except CancelledError as e:
    raise e
  except CatchableError as e:
    discard
```

markspanbroek: Yes, I actually tested all of the subtypes of JsonRpcError, but in this case we only get the base error type out of it.

I did not want to do matching on the error message string, because there is no guarantee that all node types return this exact text.

ethers/providers/jsonrpc/subscriptions.nim
```diff
@@ -152,47 +150,60 @@ proc new*(_: type JsonRpcSubscriptions,
...
-        await poll(id)
-      await sleepAsync(pollingInterval)
+  proc poll {.async: (raises: []).} =
+    try:
+      while true:
+        for id in toSeq subscriptions.callbacks.keys:
+          await poll(id)
+        await sleepAsync(pollingInterval)
+    except CancelledError:
+      discard
```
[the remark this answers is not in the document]
markspanbroek: Normally we shouldn't swallow a CancelledError, but let it bubble up. But in this case we asyncSpawn the poll() future, and that means that any Error that bubbles up becomes a Defect and crashes the process. So that's why I catch the CancelledError outside of the loop, to make sure that the loop stops, and then I do not re-raise it because there is nothing that calls poll() that needs to be cancelled, and I don't want it to bubble up to asyncSpawn.

PR #100, opened 2024-12-04
chore: fix async raises warnings

AuHau:


ethers/providers/jsonrpc/subscriptions.nim
```diff
@@ -24,6 +24,14 @@ type
 
 {.push raises:[].}
 
+template convertErrorsToSubscriptionError(body) =
+  try:
+    body
+  except CancelledError as error:
+    raise error
+  except CatchableError as error:
```
dryajov, line 32: Why are we blindly converting any exception into a `SubscriptionError`, this obfuscates the actual error.

dryajov (changes requested):
I think we need to rethink how we're handling this. Can you describe what the actual issue is?

ethers/providers/jsonrpc/subscriptions.nim
AuHau: I am just wrapping it, so it does not obfuscate it, you will get the original stack error if you ask for it. This is because json_rpc can throw (eq. any `eth_*`) `CatchableError`.

I just wanted to remove some warnings around async raises, I am happy to simply remove those completely 🤷‍♂️ Or do whatever you want, I just don't want to spend any big time on this, please.

dryajov: > I just wanted to remove some warnings around async raises

What you're doing is hiding a potential underlying problem, not properly fixing it.

I was taking a closer look at the code and I'm not even sure where `toErr` comes from and what does it do. Do you?

AuHau: > Do you?

Yes, it is coming from here https://github.com/codex-storage/nim-ethers/blob/04d35485534d99966b07394a0de96a8c70c19e87/ethers/errors.nim#L13 and as I said it wraps the error not "discards" it.

AuHau: This is just a way to "unify" the error coming from this module (Subscription) in order to communicate that these errors might come from this module.

emizzle: @dryajov I agree it's good not to overuse something like this, however calls to these RPC endpoints result in a generic `JsonRpcError` error regardless of what the error is. Other than RPC calls, there are very few exception raising paths. So this template is convenient for converting these errors to a module-level error. 

@AuHau could you catch `JsonRpcError` specifically (instead of `CatchableError`) and see if it doesn't leak any other errors?  There might be a KeyError here or there that could be caught separately, but I don't see it as really being of much benefit.

emizzle (commented):
Nice work, thanks for doing this! I don't want to drag out this PR too much, but would be nice to see `unsubscribe` not raise any exceptions.

ethers/providers/jsonrpc/subscriptions.nim
AuHau: If I change the `CatchableError` to `JsonRpcError`, the compiler will complain with this:

```
.../nim-ethers/ethers/providers/jsonrpc/subscriptions.nim(123, 40) Error: cast[type(eth_subscribe(subscriptions.client, "newHeads"))](chronosInternalRetFuture.internalChild).internalError can raise an unlisted exception: ref CatchableError
```

So this confirms what I was saying earlier - any RPC's `eth_*` call we have will _raise_ CatchableError and that is what we need to work with. So I see 

1. We add `CatchableError` to the async raises.
2. We at least wrap it to `SubscriptionError` and raise that.
3. We remove `raises` all together.
4. We do nothing.

I can't think of anything else. So @dryajov please pick. My favorite is the current approach - number 2.

dryajov: Ok, I see that we're wrapping the underlying error, that is good enough.

dryajov (approved):
LGTM

ethers/providers/jsonrpc/subscriptions.nim
emizzle: I know we discussed this already, but just for posterity, it seems that the compiler always reports that a CatchableError is leaking, and never reports the derived type. As an example, there could be a KeyError being leaked, which is derived from CatchableError, but the compiler will still report that a CatchableError is leaking.

PR #101, opened 2024-12-10
add getter to BlockTag to access the block number

marcinczenko:
Adds a getter to access the block number for BlockTag with number. Returns `Option[BlockTag]`:

```nim
let blockTag = BlockTag.init(42.u256)
if blockNumber =? blockTag.number:
  # use block number
```


AuHau (approved):
LGTM, thanks! 👍

PR #102, opened 2024-12-13
fix(jsonrpc): propagate cancelled error in call

emizzle:
- CancelledError was being swallowed by except CatchableError

### DIFF
--- a/ethers/providers/jsonrpc.nim
+++ b/ethers/providers/jsonrpc.nim
@@ -88,10 +88,16 @@ proc new*(
 proc callImpl(
   client: RpcClient,
   call: string,
-  args: JsonNode): Future[JsonNode] {.async: (raises: [JsonRpcProviderError]).} =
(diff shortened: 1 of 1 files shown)


ethers/providers/jsonrpc.nim
```diff
@@ -88,10 +88,16 @@ proc new*(
...
-  args: JsonNode): Future[JsonNode] {.async: (raises: [JsonRpcProviderError]).} =
+  args: JsonNode): Future[JsonNode] {.async: (raises: [CancelledError, JsonRpcProviderError]).} =
+
+  var response: JsonString
+  try:
+    response = await client.call(call, %args)
+  except CancelledError as e:
+    raise e
+  except CatchableError as e:
+    raiseJsonRpcProviderError e.msg
```
AuHau, line 99: I would suggest properly wrap the error with keeping not only parent's message but the whole parent error in order to keep the stacktrace.

AuHau (approved):
I will approve, but I would suggest doing proper error wrapping to persist the stack trace.

emizzle:
Closing in favour of #105

PR #105, opened 2025-02-12
fix(async): propagate async cancellation

emizzle:
Properly propagates cancellations in async procs.

### DIFF
--- a/ethers/contract.nim
+++ b/ethers/contract.nim
@@ -42,7 +42,6 @@ type
     response*: ?TransactionResponse
     convert*: ConvertCustomErrors
   EventHandler*[E: Event] = proc(event: ?!E) {.gcsafe, raises:[].}
-  ContractError* = object of EthersError
 
 func new*(ContractType: type Contract,
           address: Address,
--- a/ethers/contract.nim
+++ b/ethers/contract.nim
@@ -88,19 +88,20 @@ proc decodeResponse(T: type, bytes: seq[byte]): T =
     raiseContractError "unable to decode return value as " & $T
   return decoded
 
-proc call(provider: Provider,
-          transaction: Transaction,
-          overrides: TransactionOverrides): Future[seq[byte]] {.async: (raises: [ProviderError]).} =
-  if overrides of CallOverrides and
-     blockTag =? CallOverrides(overrides).blockTag:
+proc call(
+    provider: Provider, transaction: Transaction, overrides: TransactionOverrides
+): Future[seq[byte]] {.async: (raises: [ProviderError, CancelledError]).} =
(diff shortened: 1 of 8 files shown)


ethers/providers/jsonrpc.nim
```diff
@@ -86,29 +90,29 @@ proc new*(
   return JsonRpcProvider(client: awaitClient(), subscriptions: awaitSubscriptions())
 
 proc callImpl(
-  client: RpcClient,
-  call: string,
-  args: JsonNode): Future[JsonNode] {.async: (raises: [JsonRpcProviderError]).} =
-
-  without response =? (await client.call(call, %args)).catch, error:
+    client: RpcClient, call: string, args: JsonNode
+): Future[JsonNode] {.async: (raises: [JsonRpcProviderError, CancelledError]).} =
```
2-towns, line 94: Why did you change the `without` statement to a `try catch` ?

ethers/contract.nim
```diff
@@ -263,10 +268,23 @@ func addAsyncPragma(procedure: var NimNode) =
   let pragmas = procedure[4]
   if pragmas.kind == nnkEmpty:
     procedure[4] = newNimNode(nnkPragma)
-  procedure[4].add ident("async")
-
-macro contract*(procedure: untyped{nkProcDef|nkMethodDef}): untyped =
+  procedure[4].add nnkExprColonExpr.newTree(
```
2-towns, line 271: Here I assume that you are listing the list of possible errors that could be raised when building the contract macro .

2-towns (approved):

ethers/providers/jsonrpc.nim
emizzle: Using `catch` with `await` unfortunately swallows CancelledError. I explained it in a bit more detail here: https://discord.com/channels/895609329053474826/1168926478180810903/1339174136290873356

ethers/contract.nim
emizzle: Yes, that's correct. I don't like raising CatchableError, because that can quickly spread everywhere and makes it impossible to understand which procs are raising which exception types and then that makes it difficult to disambiguate exception types when needed. However, it's very hard to understand where this CatchableError is coming from and even harder to modify the macro to convert it.

emizzle: I managed to get rid of the CatchableError being raised. It was being raised in `send`, which was causing the compiler to report that it could leak out of the contract function. The changes are in this commit: https://github.com/codex-storage/nim-ethers/pull/105/commits/6c6aadcfc91b9c54f735e1a66301861f3821ee50

2-towns: Nice :+1:

PR #107, opened 2025-02-14
fix(ci): re-add 1.6.20 support

emizzle:
No need to drop it, yet...

The other alternative is to drop support of 1.6.x and bump a major version (2.0.0).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,7 +13,7 @@ jobs:
     timeout-minutes: 30
     strategy:
       matrix:
-        nim: [2.0.14]
+        nim: [1.6.20, 2.0.14]
     steps:
     - name: Checkout
       uses: actions/checkout@v4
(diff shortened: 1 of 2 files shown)


emizzle:
CI is failing to due missing support of the nimble option `--maximumTaggedVersions`, which means if we want to support 1.6.x, we'll need to install a nimble 0.16.4 binary in CI, and instruct our users to do the same...

It might be easier to support 2.x instead. Thoughts?

2-towns:
> CI is failing to due missing support of the nimble option `--maximumTaggedVersions`, which means if we want to support 1.6.x, we'll need to install a nimble 0.16.4 binary in CI, and instruct our users to do the same...
> 
> It might be easier to support 2.x instead. Thoughts?

I would go supporting 2.x only. 

But if we want to keep the compatibility with 1.6, forcing to use nimble 0.16.4 is totally fine and it's not too hard to setup.

AuHau:
Yeah, I would also not mind dropping 1.6. Honestly, not sure if anybody else really uses `nim-ethers` so should not really matter that much.

markspanbroek:
I'd also drop it and release a new major version.

markspanbroek:
New major version arrived: https://github.com/codex-storage/nim-ethers/releases/tag/2.0.0, closing this PR.

PR #112, opened 2025-03-19
chore: move logFilters to JsonRpcSubscriptions

2-towns:


Readme.md
```diff
@@ -204,6 +204,13 @@ This library ships with some optional modules that provides convenience utilitie
 
 - `ethers/erc20` module provides you with ERC20 token implementation and its events
 
+Workaround
+---------
+
+If you're working with Hardhat, you might encounter an issue where [subscriptions stop working after 5 minutes](https://github.com/NomicFoundation/hardhat/issues/2053).
```
AuHau, line 210:
```suggestion
If you're working with Hardhat, you might encounter an issue where [websocket subscriptions stop working after 5 minutes](https://github.com/NomicFoundation/hardhat/issues/2053).
```

ethers/providers/jsonrpc/subscriptions.nim
```diff
@@ -19,11 +19,38 @@ type
     client: RpcClient
     callbacks: Table[JsonNode, SubscriptionCallback]
     methodHandlers: Table[string, MethodHandler]
+    # We need to keep around the filters that are used to create log filters on the RPC node
+    # as there might be a time when they need to be recreated as RPC node might prune/forget
+    # about them
+    # This is used of resubscribe all the subscriptions when using websocket with hardhat
+    logFilters: Table[JsonNode, EventFilter]
```
AuHau, line 26: Why is this defined on `JsonRpcSubscriptions` level? I mean the workaround is for Websockets so IMHO it should be placed on the `WebSocketSubscriptions` level no?

AuHau (commented):
Generally good approach, but I have a question around where some things are placed, so not approving just yet.

ethers/providers/jsonrpc/subscriptions.nim
2-towns: `logFilters` was already present in `PollingSubscriptions`. So instead of duplicating the field in `WebSocketSubscriptions`, I moved it to `JsonRpcSubscriptions` so it can be used by both.

Readme.md
```diff
@@ -204,6 +204,13 @@ This library ships with some optional modules that provides convenience utilitie
 
 - `ethers/erc20` module provides you with ERC20 token implementation and its events
 
+Workaround
```
emizzle, line 207: Maybe `Hardhat websockets workaround` might be a bit more descriptive?

ethers/providers/jsonrpc/subscriptions.nim
```diff
@@ -84,6 +110,10 @@ method close*(subscriptions: JsonRpcSubscriptions) {.async: (raises: [Subscripti
   for id in ids:
     await subscriptions.unsubscribe(id)
 
+  when defined(ws_resubscribe):
+    if not subscriptions.resubscribeFut.isNil:
+      await subscriptions.resubscribeFut.cancelAndWait()
```
emizzle, line 115: Since this only applies to websocket subscriptions, we wouldn't want this called if someone compiled with `ws_resubscribe` but connected with http (and used PollingSubscriptions). Actually it wouldn't matter as `resubscribeFut` would be nil and we do a nil check. 

We could probably create an overload for `WebsocketSubscriptions.close` instead of putting this in the base class `close`:

```nim
method close*(subscriptions: WebsocketSubscriptions) {.async.} =
  await procCall JsonRpcSubscriptions(subscriptions).close()
  if not subscriptions.resubscribeFut.isNil:
      await subscriptions.resubscribeFut.cancelAndWait()
```

ethers/providers/jsonrpc/subscriptions.nim
```diff
@@ -19,11 +19,38 @@ type
     client: RpcClient
     callbacks: Table[JsonNode, SubscriptionCallback]
     methodHandlers: Table[string, MethodHandler]
+    # We need to keep around the filters that are used to create log filters on the RPC node
+    # as there might be a time when they need to be recreated as RPC node might prune/forget
+    # about them
+    # This is used of resubscribe all the subscriptions when using websocket with hardhat
```
emizzle, line 25:
```suggestion
    # Used by both PollingSubscriptions and WebsocketSubscriptions to store
    # subscription filters so the subscriptions can be recreated. With
    # PollingSubscriptions, the RPC node might prune/forget about them, and with
    # WebsocketSubscriptions, when using hardhat, subscriptions are dropped after 5
    # minutes.
```

config.nims
```diff
@@ -8,3 +8,4 @@ when fileExists("nimble.paths"):
 
 when (NimMajor, NimMinor) >= (2, 0):
   --mm:refc
+  --define:ws_resubscribe
```
emizzle, line 11: If we include this, then ethers will always be compiled with the resubscription feature enabled. Shouldn't we leave this to application implementations to enable this flag?

ethers/providers/jsonrpc/subscriptions.nim
```diff
@@ -19,11 +19,38 @@ type
...
 
 {.push raises:[].}
 
+when defined(ws_resubscribe):
+  # This is a workaround to manage the 5 minutes limit due to hardhat.
+  # See https://github.com/NomicFoundation/hardhat/issues/2053#issuecomment-1061374064
+  proc resubscribeWebsocketEventsOnTimeout*(subscriptions: JsonRpcSubscriptions) {.async.} =
+    while true:
+      await sleepAsync(4.int64.minutes)
+      for id, callback in subscriptions.callbacks:
```
emizzle, line 40: We either need an `AsyncLock` or a copy because if subscriptions are created or deleted during this routine, there will be an error like "collection cannot be modified while iterating". 

We should probably add a test for this scenario if possible as well.

emizzle (commented):
Looking good. Left some comments.

ethers/providers/jsonrpc/subscriptions.nim
```diff
@@ -19,11 +19,38 @@ type
...
+      resubscribeFut: Future[void]
   MethodHandler* = proc (j: JsonNode) {.gcsafe, raises: [].}
   SubscriptionCallback = proc(id: JsonNode, arguments: ?!JsonNode) {.gcsafe, raises:[].}
 
 {.push raises:[].}
 
+when defined(ws_resubscribe):
+  # This is a workaround to manage the 5 minutes limit due to hardhat.
+  # See https://github.com/NomicFoundation/hardhat/issues/2053#issuecomment-1061374064
+  proc resubscribeWebsocketEventsOnTimeout*(subscriptions: JsonRpcSubscriptions) {.async.} =
```
emizzle, line 37: May as well annotate with `{.async: (raises: [...]).}`, no?

ethers/providers/jsonrpc/subscriptions.nim
AuHau: Yeah, that makes sense - thanks for the explanation! 👍

ethers/providers/jsonrpc/subscriptions.nim
2-towns: Yes, I wanted to do that, but I didn’t know how to call the 'parent method" in Nim.

Thanks I will apply it.

config.nims
2-towns: Yes definitely. I added the flag to make sure everything was compiling but I forgot to remove it :flushed:.

ethers/providers/jsonrpc/subscriptions.nim
2-towns: Very close. I wrote a test and got ` Unhandled exception: /home/arnaud/.choosenim/toolchains/nim-2.0.14/lib/pure/collections/tables.nim(699, 13) `len(t) == L` the length of the table changed while iterating over it`. 

Okay I am on it.

2-towns: Okay so I added a test handling this case but to do that, I needed to add a value to the flag. So by default, it is 240 seconds (4 minutes). And you can change the value using `--define:ws_resubscribe=180`. So for the tests, I am using `--define:ws_resubscribe=3` to quickly trigger the re-subscriptions. I think it is great because it is giving more flexibility for the re-subscription period.

testmodule/providers/jsonrpc/testWsResubscription.nim
```diff
@@ -0,0 +1,55 @@
...
+    let emptyHandler = proc(log: ?!Log) = discard
+    let id = await subscriptions.subscribeLogs(filter, emptyHandler)
+
+    check id in subscriptions.logFilters
+    check subscriptions.logFilters.len == 1
+
+    await sleepAsync(4.int64.seconds)
+
+    # The previous subscription should not be in the log filters
+    check not (id in subscriptions.logFilters)
```
emizzle, line 52:
```suggestion
    check id notin subscriptions.logFilters
```

testmodule/providers/jsonrpc/testWsResubscription.nim
```diff
@@ -0,0 +1,55 @@
...
+
+  test "resubscribe events take effect with new subscription IDs in the log filters":
+    let filter = EventFilter(address: Address.example, topics: @[array[32, byte].example])
+    let emptyHandler = proc(log: ?!Log) = discard
+    let id = await subscriptions.subscribeLogs(filter, emptyHandler)
+
+    check id in subscriptions.logFilters
+    check subscriptions.logFilters.len == 1
+
+    await sleepAsync(4.int64.seconds)
```
emizzle, line 49: Instead of hardcoding values here, I think we should use the value from the `-d` cli param instead, or even better is to pass that value into the WebSocketSubscription constructor so the value in the tests can be set accordingly, defaulting to the value set in -d, which itself defaults to 240 (4 mins). 

```nim
type
  WebSocketSubscriptions = ref object of JsonRpcSubscriptions
    resubscribeInterval: int

proc new*(_: type JsonRpcSubscriptions,
          client: RpcWebSocketClient,
          resubscribeInterval = wsResubscribe): JsonRpcSubscriptions =
  let subscriptions = WebSocketSubscriptions(client: client, resubscribeInterval: resubscribeInterval)
  # ..
  subscriptions
```

Then in the ws resubscription tests:
```nim
  setup:
    client = newRpcWebSocketClient()
    await client.connect("ws://"  & getEnv("ETHERS_TEST_PROVIDER", "localhost:8545"))
    subscriptions = JsonRpcSubscriptions.new(client, 3)
```

testmodule/test.nimble
```diff
@@ -8,3 +8,7 @@ requires "asynctest >= 0.4.0 & < 0.5.0"
 task test, "Run the test suite":
   exec "nimble install -d -y"
   exec "nim c -r test"
+
+task testWsResubscription, "Run the test suite":
+  exec "nimble install -d -y"
+  exec "nim c --define:ws_resubscribe=3 -r providers/jsonrpc/testWsResubscription"
```
emizzle, line 14: If we avoid hardcoding the values in tests and instead pass the resubscribe interval value into the constructor, we can avoid having separate tests, and having to re-run nimble install (which is quite slow).

testmodule/providers/jsonrpc/testWsResubscription.nim
emizzle: And can change the sleep timeout (line 204) to `sleepAsync(subscriptions.resubscribeInterval)`

ethers/providers/jsonrpc/subscriptions.nim
```diff
@@ -137,34 +167,69 @@ method subscribeLogs(subscriptions: WebSocketSubscriptions,
...
+  await procCall JsonRpcSubscriptions(subscriptions).close()
+  when defined(ws_resubscribe):
+    if not subscriptions.resubscribeFut.isNil:
+        await subscriptions.resubscribeFut.cancelAndWait()
+
+when defined(ws_resubscribe):
+  # This is a workaround to manage the 5 minutes limit due to hardhat.
+  # See https://github.com/NomicFoundation/hardhat/issues/2053#issuecomment-1061374064
+  proc resubscribeWebsocketEventsOnTimeout*(subscriptions: WebsocketSubscriptions) {.async: (raises: [CancelledError]).} =
+    while true:
```
emizzle, line 204: Maybe we should do a bit of validation here, returning if the resubscribe interval is <= 0?

ethers/providers/jsonrpc/subscriptions.nim
```diff
@@ -119,10 +144,15 @@ method subscribeBlocks(subscriptions: WebSocketSubscriptions,
...
-    return id
+  try:
+    withLock(subscriptions):
+      convertErrorsToSubscriptionError:
+        let id = await subscriptions.client.eth_subscribe("newHeads")
+        subscriptions.callbacks[id] = callback
+        return id
+  except AsyncLockError as e:
+     error "Lock error when trying to subscribe to blocks", err = e.msg
+     raise newException(SubscriptionError, "Cannot subscribe to the blocks because of lock error")
```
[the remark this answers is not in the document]
2-towns: Yes I was hesitating between what I did and this approach, but yes I agree, your suggestion is better.

testmodule/providers/jsonrpc/testWsResubscription.nim
```diff
@@ -0,0 +1,55 @@
...
+
+    let subscription = await subscriptions.subscribeLogs(filter, emptyHandler)
+
+    # Wait until the re-subscription starts
+    await sleepAsync(3.int64.seconds)
+
+    try:
+        await subscriptions.unsubscribe(subscription)
+    except CatchableError:
+        fail()
```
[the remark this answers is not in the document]
2-towns: Yep. I was testing locally and with one subscription, the test was failing without lock. But I agree it's better to test with multiple subscriptions.

testmodule/test.nimble
2-towns: Okay, so with `when defined(ws_resubscribe)`, we’d still need to pass the `--define:ws_resubscribe=0` flag during compilation. We’d set it to `0` to disable re-subscription by default for other tests. However, this approach introduces two problems:

1. It’s not an elegant solution.
2. We would test only the code with the `ws_resubscribe` flag but not without it

Instead of relying on `when defined(ws_resubscribe)`, I propose setting the default value of the flag to `0`. This way, re-subscription is disabled by default and only activated if a compilation flag like `--define:ws_resubscribe=180` is provided. This allows us to easily test both scenarios: we can pass the value explicitly in `JsonRpcSubscriptions.new` to enable re-subscription and test it, while the default `0` ensures it’s off for other cases.

emizzle: Yes, good idea. I think it better to have a default value of 0 anyway, since this issue affects local instances of Hardhat only.

> We would test only the code with the ws_resubscribe flag but not without it
Yes, but that's ok, because the value in the flag is _only used_ to pass into the constructor. And so we can vary our tests by changing the value passed in the constructor (with `0` as one of the tests).

Mainly I'd like to avoid having two separate test runs with two separate `nimble install` calls.

ethers/providers/jsonrpc/subscriptions.nim
```diff
@@ -138,33 +194,36 @@ method subscribeLogs(subscriptions: WebSocketSubscriptions,
...
+    withLock(subscriptions):
+      subscriptions.callbacks.del(id)
+      discard await subscriptions.client.eth_unsubscribe(id)
   except CancelledError as e:
     raise e
   except CatchableError:
     # Ignore if uninstallation of the subscribiton fails.
     discard
 
+method close*(subscriptions: WebSocketSubscriptions) {.async: (raises: [CancelledError, SubscriptionError]).} =
```
emizzle, line 216: We should probably avoid raising any exceptions in our closing routines, including `CancelledError`. Closing routines are likely going to be called in their own defer blocks or closing sequences after cancellations have been handled and things are being cleaned up.

emizzle (approved):
Nice work, Arnaud. Thanks for making all the changes. A couple of suggestions, but nothing major and blocking, so will approve now 👍

markspanbroek (approved):
Good stuff @2-towns!

The only thing that I'm left wondering with is whether the locking won't give us trouble down the road; if resubscribing takes a long time then it locks up subscriptions pretty bad. But since we're only using this with Hardhat that won't be much of a problem.

ethers/providers/jsonrpc/subscriptions.nim
2-towns: Okay done

ethers/providers/jsonrpc/subscriptions.nim
```diff
@@ -93,18 +98,68 @@ proc getCallback(subscriptions: JsonRpcSubscriptions,
...
-  let subscriptions = WebSocketSubscriptions(client: client)
   proc subscriptionHandler(arguments: JsonNode) {.raises:[].} =
     let id = arguments{"subscription"} or newJString("")
     if callback =? subscriptions.getCallback(id):
       callback(id, success(arguments))
   subscriptions.setMethodHandler("eth_subscription", subscriptionHandler)
+
+  if resubscribeInterval > 0:
+    if resubscribeInterval >= 300:
+      info "Resubscription interval greater than 300 seconds is useless for hardhat workaround", resubscribeInterval = resubscribeInterval
```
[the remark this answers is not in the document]
2-towns: Done

PR #113, opened 2025-04-01
chore: add EIP-1559 implementation for gas price

2-towns:


arnetheduck:
see also https://github.com/status-im/nim-eth/blob/master/eth/eip1559.nim

ethers/providers/jsonrpc.nim
```diff
@@ -151,6 +154,12 @@ method getGasPrice*(
     let client = await provider.client
     return await client.eth_gasPrice()
 
+method getMaxPriorityFeePerGas*(
+    provider: JsonRpcProvider
+): Future[UInt256] {.async: (raises: [ProviderError, CancelledError]).} =
```
emizzle: Why does this need to be `async`?

emizzle (commented):
Generally LGTM. Should we include:
1. Validation of `maxPriorityFeePerGas`. Maybe a warning if it's too low (< 500k) that the fee is low and it reduces the chances of the transaction being included in the block
2. Tests on bounds of `maxPriorityFeePerGas`, so that we know what bounds are acceptable for users.

ethers/providers/jsonrpc.nim
2-towns: Because the maxPriorityFeePerGas can be retrieved by making a call to `eth_maxPriorityFeePerGas` when the endpoint is available.

2-towns: And it seems to be supported in Linea. So maybe we should the same strategy as here: https://github.com/NomicFoundation/hardhat/blob/b914c6c00f1ef8e47a895bb75a38b766882d792c/packages/hardhat-ignition-core/src/internal/execution/jsonrpc-client.ts#L739.

emizzle: Adding a sane default is not a bad idea, though ethers doesn't really have a concept of config. So maybe we could hardcode a fallback value?

2-towns: Yes you can look at my [last commit](https://github.com/codex-storage/nim-ethers/pull/113/commits/228d18efccca9211dab5d0568e23726fc66b02f4). If the call to `eth_maxPriorityFeePerGas` fails, it will return `provider.maxPriorityFeePerGas` which has a default value and can be defined (if needed) when creating the provider.

ethers/signer.nim
```diff
@@ -124,8 +130,24 @@ method populateTransaction*(
...
+
+  let blk = await signer.provider.getBlock(BlockTag.latest)
+
+  if baseFeePerGas =? blk.?baseFeePerGas:
+    let maxPriorityFeePerGas = transaction.maxPriorityFeePerGas |? (await signer.provider.getMaxPriorityFeePerGas())
+    populated.maxPriorityFeePerGas = some(maxPriorityFeePerGas)
+
+    # Multiply by 2 because during times of congestion, baseFeePerGas can increase by 12.5% per block.
+    # https://github.com/ethers-io/ethers.js/discussions/3601#discussioncomment-4461273
+    let maxFeePerGas = transaction.maxFeePerGas |? baseFeePerGas * 2 + maxPriorityFeePerGas
```
emizzle:
```suggestion
    let maxFeePerGas = (transaction.maxFeePerGas |? baseFeePerGas) * 2 + maxPriorityFeePerGas
```
Just a suggestion for better readability

ethers/signer.nim
```diff
@@ -124,8 +130,24 @@ method populateTransaction*(
...
+    populated.maxFeePerGas = some(maxFeePerGas)
+
+    populated.gasPrice = none(UInt256)
+
+    trace "EIP-1559 is supported", maxPriorityFeePerGas = maxPriorityFeePerGas, maxFeePerGas = maxFeePerGas
+  else:
+    populated.gasPrice = some(transaction.gasPrice |? (await signer.getGasPrice()))
+    trace "EIP-1559 is not supported", gasPrice = populated.gasPrice
```
emizzle, line 152: Should we set `populated.maxFeePerGas = none(UInt256)` just in case here? Thinking of a case where the transaction.maxFeePerGas was already populated but the block didn't have a `baseFeePerGas` value.

ethers/providers/jsonrpc.nim
```diff
@@ -151,6 +154,18 @@ method getGasPrice*(
...
     return await client.eth_gasPrice()
 
+method getMaxPriorityFeePerGas*(
+    provider: JsonRpcProvider
+): Future[UInt256] {.async: (raises: [CancelledError]).} =
+    try:
+      convertError:
+        let client = await provider.client
+        return await client.eth_maxPriorityFeePerGas()
+    except ProviderError:
```
emizzle: `eth_maxPriorityFeePerGas` is generated by a `nim-json-rpc` macro and it will not raise a `ProviderError`. Likely you could catch a `JsonRpcError` here instead.

emizzle (approved):
Nice, looking good. One minor request, a question, and a minor suggestion :) I'll approve now, and assume the minor request is updated (`ProviderError` -> `JsonRpcError`)

ethers/signer.nim
2-towns: It's `let maxFeePerGas = transaction.maxFeePerGas |? (baseFeePerGas * 2 + maxPriorityFeePerGas)`, the `transaction.maxFeePerGas` is an override when you want to set manually the value of `maxFeePerGas` without extra calculation. 

I applied the change.

ethers/signer.nim
2-towns: I added it just in case, but it should not have any impact. If the block doesn't have `baseFeePerGas`  that means EIP-1559 is not supported so `maxFeePerGas` will not be used.

ethers/providers/jsonrpc.nim
2-towns: It is under the template `convertError` so `JsonRpcError` will be transformed to `JsonRpcProviderError`. I replaced the `ProviderError` to `JsonRpcProviderError` in order to be more precise.

emizzle: Ah true, good point. Sorry I missed that. 👍

ethers/signer.nim
emizzle: Fair point. When the transaction is signed in the wallet, the first condition checked is whether or not `transaction.maxFeePerGas` has a value (see line 35 in `signers/wallet/signing.nim`), which ends up setting `signable.txType = TxEip1559` and could have side effects later on.

emizzle: We may also want to set `maxPriorityFeePerGas` to `none` just in case. Just a suggestion.

2-towns: Yes you are right. I just updated the PR to set none to maxPriorityFeePerGas to avoid side effect. Thanks.

PR #115, opened 2025-04-09
Support gas estimation of contract calls

markspanbroek:
Supports the following syntax:

```nim
let gas = await token.estimateGas.mint(account, amount)
```

Consists of the following changes:
- refactor of the `contract` module that splits it into several smaller modules (45964f5d6c2fdd626d288d97065008e8a9af5807, 55ab97fb7b4a1788de05e61eeae8c165310ac947)
- the `estimateGas` syntax (5586ee2697af4ee56fa4dc89af8f68942939aefb, c4f1d7529c29b5a5ebe5dbafc2c365b1989f0620, 795bf0eb4d62209bc0941939873fb3b4525b1f01)

~~Depends on #114~~


testmodule/testGasEstimation.nim
```diff
@@ -41,3 +41,37 @@ suite "gas estimation":
...
+    # ensure that time of latest block and pending block differ
+    check (await contract.getTime(overrides=latest)) != time
+
+    # estimate gas
+    let gas = await contract.estimateGas.checkTimeEquals(time)
+    let overrides = TransactionOverrides(gasLimit: some gas)
+
+    # fails with "Transaction ran out of gas" when gas estimation
+    # was not done using the pending block
+    await contract.checkTimeEquals(time, overrides)
```
emizzle, line 61: This is quite confusing to me. The "pending" block is the next block that will be produced and because Hardhat has automine enabled, the next block is produced immediately when a transaction occurs. Therefore, a transaction happens and block should be produced when `checkTimeEquals` is called, so wouldn't that make the `block.timestamp` of the transaction equal to the "pending" block timestamp just before it was called?

emizzle (approved):
Looks great, Mark. It's a bit difficult to distinguish new changes since there was also a lot of reorganisation, but I really like how the contract production was broken up.

AuHau (approved):
LGTM, but agree with Eric that it is rather hard to see the changes.

markspanbroek:
> agree with Eric that it is rather hard to see the changes.

Yes, I apologize. I should have probably broken this up into two PRs. It started out with just 2 commits and I thought; you can just look at the two commits separately. But then I had to add 3 more commits and at that point I should have probably split it into two PRs.

testmodule/testGasEstimation.nim
markspanbroek: Updated the comments in 6ef1ce5300b7bac35ccd89a5d1792300f5d6e10e so they are less confusing, I hope.

PR #122, opened 2025-09-10
refactor!: new implementation of subscriptions

markspanbroek:
New implementation of subscriptions, that keeps a local view of all subscriptions instead of letting the JSON-RPC node handle the subscriptions.

Main advantages:
- falls back to polling if websocket updates no longer happen
- no need to resubscribe when something goes wrong with the connection
- reduces the number of RPC calls because we now check the block's bloom filter before getting logs

Additionally:
- `JsonRpcProvider.new()` is replaced by `JsonRpcProvider.connect()`, to get an error when the connection fails
- subscription callbacks no longer get a `Result` type, because subscription errors are now handled elsewhere


AuHau:
Hey @markspanbroek!

Thanks for the PR! I will try to have a look by the end of the week.

arnetheduck:
fyi, json-rpc has gained support for bidirectional comms and will be deprecating the existing message hook: https://github.com/status-im/nim-json-rpc/pull/249 in an upcoming release. It's an open question whether the hook will be replaced, but likely not, since subscriptions can (and should) use the rpc router instead. Possibly there might be a hook for already-decoded messages _or_ post-router lookup for messages that don't match the router, depending on demand.

In the new setup, notifications are handled via the regular `RpcRouter`.

